PR #4, opened 2018-07-22
Adds per-environment configs.

ettomatic:


config/config.exs
```diff
@@ -27,4 +27,6 @@ use Mix.Config
 # Configuration from the imported file will override the ones defined
 # here (which is why it is important to import them last).
 #
-#     import_config "#{Mix.env}.exs"
+config :fetcher, :environment, Mix.env()
```
samfrench: I read last week Mix is not available in production. If this is the case, If we use Application.get_env then we'd have to set it somewhere...

I prefer this way if it is available as it's much simpler.

PR #5, opened 2018-07-24
Adds a Dockerfile.

ettomatic:
~STILL WIP~

Dockerfile and Makefile to allow a more automated way to build this application.


Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM qixxit/elixir-centos
+
+MAINTAINER Ettore Berardi <ettore.berardi@bbc.co.uk>
+
+LABEL uk.co.bbci.api.mozart="true"
```
samfrench: I think we should have this as: uk.co.bbc.news or mozart, but bbci seems unusual.

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+#ADD dev.pem /opt/app/dev.pem
```
samfrench: ^

samfrench:
Very good. I'm going to create a CentOS7 and Elixir image tomorrow unless you wanted to, or to pair. I'm also going to read up on what we're producing to see if there are any flags we should set when compiling. At the moment I don't know what it's doing to generate the output.

PR #9, opened 2019-02-19
Move away from mock and create a fake origin serving components

woodyblah:
I noticed the component_test.exs specs were failing for me, with a very strange error, to do with meck. As I couldn't work out the cause I decided it might be better practice to move away from using Mock anyway as this seems not to be nest practice in Elixir

Based on [this article](https://medium.com/flatiron-labs/rolling-your-own-mock-server-for-testing-in-elixir-2cdb5ccdd1a0) I have made an attempt at creating a fake origin that serves components, that only starts up when the tests run. This decouples our tests from HTTPoison and even our own HTTPClient but I have yet to find a nicer way of simulating timeouts other than simply putting the Fake Origin to sleep.


lib/mozart_fetcher/fake_origin.ex
```diff
@@ -0,0 +1,18 @@
...
+
+  plug :match
+  plug :dispatch
+
+  get "/success" do
+    send_resp(conn, 200, ~s({"head":[],"bodyInline":"<DIV id=\\"site-container\\">","bodyLast":[]}) )
+  end
+
+  get "/timeout" do
+    :timer.sleep(3100)
```
ettomatic: we could have a very short timeout for the test env right?

woodyblah: I think that might be the way to go yeah. I'll see how easy this is to implement. I hope just a small thing in application.ex

PR #10, opened 2019-02-28
Add metrics, run formatter, update dependencies & remove unused Envelope.build function

james-bowers:
Metrics appear in cloudwatch:

screen shot 2019-03-06 at 15 13 39

Dependant on:
- News infra branch
- ExMetrics library: https://github.com/bbc/ExMetrics


config/metrics.exs
```diff
@@ -0,0 +1,12 @@
+use Mix.Config
+
+config :ex_metrics,
+  metrics: [
```
woodyblah: 🤔 so you have to list all your metrics up front? Can't we just send whatever metric?

config/dev.exs
```diff
@@ -1,3 +1,7 @@
 use Mix.Config
 
-config :mozart_fetcher, timeout: 3000
\ No newline at end of file
+config :mozart_fetcher, timeout: 3000
+
+config :ex_metrics,
+  send_metrics: false,
+  raise_on_undefined_metrics: true
```
samfrench: Why do we need to define them? Why not just send whatever it receives?

lib/mozart_fetcher/component.ex
```diff
@@ -9,14 +9,16 @@ defmodule MozartFetcher.Component do
   end
 
   defp process(config, {:ok, %HTTPoison.Response{status_code: 200, body: body}}) do
-    %Component{index: 0, id: config.id, status: 200, envelope: Envelope.build({:ok, body})}
+    ExMetrics.increment("success.component.process")
+    %Component{index: 0, id: config.id, status: 200, envelope: Envelope.build(body)}
   end
 
   defp process(_config, {:error, %HTTPoison.Error{reason: reason}}) do
+    ExMetrics.increment("error.component.process")
```
samfrench: This is a bit off topic from this PR. If we're making this compatible with requester then we should return components back with non 200 statuses which are likely to be empty envelopes. This error matcher doesn't do this and also the success one above hardcodes the 200.

config/metrics.exs
james-bowers: yes, so this is what I was going to discuss with people. This is an optional feature.

This was a step towards automating our alarms and dashboards for any app that uses the ExMetrics library. It also opens the door to other potential features:

The pros are:
1. All metrics are clearly defined in a single place

2. Avoid spelling errors in metrics

3. Encourages more discipline in metric naming - poorly named metrics stand out

4. We can make use of the `mix list_metrics` command the ExMetrics library provides

5. We can create more mix commands in the `ExMetrics` library - for example it is possible that using this list of metrics with other defined information, we could generate cloudformation for alarms, and/or grafana dashboards.

6. If in the future we add optional thresholds for metrics, in the metrics config file, we could also specify a callback module that performs an action when a metric reaches the threshold

7. These checks are only done in `test` and `dev` mix envs, on live, a metric with any name is sent through.

It does have cons:
1. We might not want to implement the mix tasks that generate alarms and grafana charts because this is too close to the code, and is currently kept separate for reasons

woodyblah: I think the added con is a lack of dynamic metrics, i.e in requester we currently metric the component name and status code, and we don't want to specify in advance a different metric for every combination of every possible component id and status

james-bowers: well, my plan was great albeit short-lived

james-bowers: actually, it seems it wasn't even great 😆

james-bowers: Thinking about this more, I believe I discarded that kind of dynamic metric as we cannot query cloudwatch from grafana with something like "error.component.non200.*.*" so we'd have to hardcode those metrics to grafana & specifying alarms anyway, so we might as well do that here?

james-bowers: I've now changed the ExMetrics library so that we only log an error message when a metric is undefined in this config. This means that we can continue to specify the metrics we know about here and wish to use when building alarms and dashboards - but does not limit us in the test and dev environments from using dynamic metrics

https://github.com/bbc/ExMetrics/pull/5

lib/mozart_fetcher/http_client.ex
```diff
@@ -1,11 +1,15 @@
 defmodule HTTPClient do
+  use ExMetrics
+
   def get(endpoint) do
-    cert = Application.get_env(:mozart_fetcher, :dev_cert_pem)
-    timeout = Application.get_env(:mozart_fetcher, :timeout)
+    ExMetrics.timeframe "function.timing.http_client.get" do
```
BradleyTesterSpratt: is .timeframe recording how long get takes?

JoeARO: @BradleyTesterSpratt take a look at the macro here:
https://github.com/bbc/ExMetrics/blob/master/lib/ex_metrics.ex#L19
See if you can figure it out if not ill try explain it how James explained it to me :)

james-bowers: Have a little read of this: https://elixir-lang.org/getting-started/meta/macros.html

BradleyTesterSpratt: yeah, i looked at that macro, thats where my question came from. :)

PR #12, opened 2019-03-20
Implement logging

JoeARO:


lib/mozart_fetcher/envelope.ex
```diff
@@ -12,7 +13,8 @@ defmodule MozartFetcher.Envelope do
 
       {:error, _, _} ->
         ExMetrics.increment("error.envelope.decode")
+        Stump.log(:error, %{message: "Failed to decode Envelope"})
```
woodyblah: I think it might be good to also log the `envelope`

JoeARO: Done, have labelled it `body` and passed through from the `case` above

config/prod.exs
```diff
@@ -1,9 +1,17 @@
...
 
 config :ex_metrics,
   send_metrics: true,
   raise_on_undefined_metrics: false
+
+config :logger,
+  backends: [{LoggerFileBackend, :file}]
+
+config :logger, :file,
+  path: "/var/log/mozart_fetcher/error.log",
```
woodyblah: Happy with this being hard coded for now, but a future improvement will be to get this from cosmos config/environment variables

config/prod.exs
```diff
@@ -1,9 +1,17 @@
...
+  format: "$message\n",
```
james-bowers: as the format is set in the `config.exs` - I don't think we need to set it here too?

lib/mozart_fetcher/component.ex
```diff
@@ -1,5 +1,6 @@
 defmodule MozartFetcher.Component do
-  alias MozartFetcher.{Component, Config, Envelope, LocalCache}
+  alias   MozartFetcher.{Component, Config, Envelope, LocalCache}
```
james-bowers: is this from the changes of running `mix format`? I think it'd be good to keep using the format provided by `mix format`

lib/mozart_fetcher/component.ex
```diff
@@ -1,5 +1,6 @@
...
+  require Stump
```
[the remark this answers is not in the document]
JoeARO: I don't think you can do a require as, but maybe an alias here would work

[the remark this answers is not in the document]
JoeARO: My only concern is that if someone overlooks that and then thinks that we don't need the wrapper for `Logger` anymore or they think the Logger can take a map and do work with the original `Logger` not `Stump`

lib/mozart_fetcher/component.ex
JoeARO: Sure ill run that

config/prod.exs
JoeARO: Because the other one is setting the log format for the console and this one is setting it for the file backend

lib/mozart_fetcher/component.ex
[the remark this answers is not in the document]
ettomatic: ah true! wondering what's the difference using:
```
alias Stump, as: Logger
```

maybe worth a look?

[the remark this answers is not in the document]
ettomatic: btw whay we needed to require it instead of aliasing?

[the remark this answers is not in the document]
ettomatic: bit of overlapping in the comments! I see your point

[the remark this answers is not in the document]
ettomatic: still not sure why you require it instead of aliasing

config/prod.exs
james-bowers: How do we decide what should be in these "application" configs set at compile time - and what should be put in cosmos? I see the benefit of cosmos applications being that they can change config values between environments - but this log file path won't change between environments, and could be kept here in the `prod.exs` config.

woodyblah: Hmm, good point. I was thinking more from the point of view that the log path is used by more than just the application. We might need to mention the logpath in the systemd unit file, and the config for the td-agent

james-bowers: ah yes, good point

james-bowers:
hmm, I get an error running the tests locally:
```
1) test build an envelope it returns an empty Envelope if not valid content (MozartFetcher.EnvelopeTest)
     test/envelope_test.exs:19
     ** (CaseClauseError) no case clause matching: {:error, :invalid, 1}
     code: assert Envelope.build(component) == %Envelope{}
     stacktrace:
       (mozart_fetcher) lib/mozart_fetcher/envelope.ex:9: MozartFetcher.Envelope.build/1
       test/envelope_test.exs:21: (test)

```

is it just me? `master` branch passes

lib/mozart_fetcher/component.ex
```diff
@@ -1,5 +1,6 @@
 defmodule MozartFetcher.Component do
   alias MozartFetcher.{Component, Config, Envelope, LocalCache}
+  alias Stump
```
james-bowers: I don't think you need to alias this

lib/mozart_fetcher/envelope.ex
```diff
@@ -1,5 +1,6 @@
 defmodule MozartFetcher.Envelope do
   alias MozartFetcher.{Envelope}
+  require Stump
```
james-bowers: I don't think we need to require or alias this

lib/mozart_fetcher/fetcher.ex
```diff
@@ -1,10 +1,12 @@
 defmodule MozartFetcher.Fetcher do
   alias MozartFetcher.{Component}
+  require Stump
```
james-bowers: same here? 🤔

lib/mozart_fetcher/component.ex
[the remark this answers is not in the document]
james-bowers: I'm not sure we even need to alias it 🤔

PR #13, opened 2019-03-22
fix no non 200 matcher for fetching components - and add per component metric

james-bowers:


lib/mozart_fetcher/component.ex
```diff
@@ -8,13 +8,15 @@ defmodule MozartFetcher.Component do
     process(config, get(config))
   end
 
-  defp process(config, {:ok, %HTTPoison.Response{status_code: 200, body: body}}) do
+  defp process(config, {:ok, %HTTPoison.Response{status_code: status_code, body: body}}) do
```
JoeARO: We discussed this and decided that for now there best thing to have would be a success.200/202/404/500 and have it working and have future work to better represent what has happened here in the metrics names.
@james-bowers and @woodyblah feel free to elaborate more on this if you feel necessary

james-bowers: we could have both?
```
ExMetrics.increment("success.component.process.#{status_code}")
ExMetrics.increment("success.component.process.#{config.id}.#{status_code}")
```

JoeARO: I was more getting at the success and the potential of having a status code that makes us believe otherwise, if you get me? I realise maybe I only spoke about that when talking to Ryan not sure if he mentioned it when you were still in

james-bowers: ahh I see, yeah, we could have a guard that checks the status code, and so have something like:
```
def process(config, status) when status in [200..399] do
    # success
end

def process(config, status) when status in [400..599] do
    # failure
end
```

PR #15, opened 2019-04-04
Component error'ing broke Poison.decode

JoeARO:
Due to the original return from the process method if there was a
failure it would cause the Poison.decode to kill the process for that request.
Handling it this way (similar to how Requester handles it) means that it
will still be able to parse a empty envelope while letting us know that
the component has timed out or there is a server error.

### DIFF
--- a/lib/mozart_fetcher/component.ex
+++ b/lib/mozart_fetcher/component.ex
@@ -16,10 +16,18 @@ defmodule MozartFetcher.Component do
   defp process(_config, {:error, %HTTPoison.Error{reason: reason}}) do
     ExMetrics.increment("error.component.process")
     Stump.log(:error, %{message: "Failed to process HTTP request, reason: #{reason}"})
-    {:error, reason}
+    failed_component(reason)
   end
 
   defp get(config) do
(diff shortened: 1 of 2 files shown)


lib/mozart_fetcher/component.ex
```diff
@@ -16,10 +16,18 @@ defmodule MozartFetcher.Component do
...
-    {:error, reason}
+    failed_component(reason)
   end
 
   defp get(config) do
     LocalCache.get_or_store(config.endpoint, fn -> HTTPClient.get(config.endpoint) end)
   end
+
+  defp failed_component(:timeout) do
+    %Component{index: 0, id: "", status: 408, envelope: %Envelope{}}
```
james-bowers: Doesn't this mean the component id will be stripped from the debug mode view? It'd be good if we kept the component.id

woodyblah: Good spot, I think this could also effect how assembler works out if they were must_succeed or not it will probably look for the ID in the responses

JoeARO: created a PR here: https://github.com/bbc/mozart-fetcher/pull/19

PR #17, opened 2019-04-04
Implement Jason JSON package

JoeARO:
The Jason JSON package has greater performance than the Poison JSON package although with decoding there was some work around to put the data into the Struct formats that we are currently expecting in the application.

This has allowed for better error handling around the areas we use the package and we shouldn't fallover from broken component JSON but will error out instead.

We also now return a 500 if there are any issues with decoding the initial component body.


lib/mozart_fetcher/decoder.ex
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def list_to_struct_list(data, struct) do
+    case Jason.decode(data) do
+      {:ok, data} ->
+        {:ok, Enum.map(data["components"], fn x -> to_struct(x, as: struct) end)}
+      {:error, _} -> {:error}
+    end
+
+    # build list while iterating through each section of the JSON and turning it into a struct
```
samfrench: Is the list built above? If so, can we put the comments before the code they are commenting.

mix.lock
```diff
@@ -3,18 +3,19 @@
...
-  "plug": {:hex, :plug, "1.7.2", "d7b7db7fbd755e8283b6c0a50be71ec0a3d67d9213d74422d9372effc8e87fd1", [:mix], [{:mime, "~> 1.0", [hex: :mime, repo: "hexpm", optional: false]}, {:plug_crypto, "~> 1.0", [hex: :plug_crypto, repo: "hexpm", optional: false]}], "hexpm"},
+  "plug": {:hex, :plug, "1.8.0", "9d2685cb007fe5e28ed9ac27af2815bc262b7817a00929ac10f56f169f43b977", [:mix], [{:mime, "~> 1.0", [hex: :mime, repo: "hexpm", optional: false]}, {:plug_crypto, "~> 1.0", [hex: :plug_crypto, repo: "hexpm", optional: false]}, {:telemetry, "~> 0.4", [hex: :telemetry, repo: "hexpm", optional: true]}], "hexpm"},
   "plug_cowboy": {:hex, :plug_cowboy, "2.0.2", "6055f16868cc4882b24b6e1d63d2bada94fb4978413377a3b32ac16c18dffba2", [:mix], [{:cowboy, "~> 2.5", [hex: :cowboy, repo: "hexpm", optional: false]}, {:plug, "~> 1.7", [hex: :plug, repo: "hexpm", optional: false]}], "hexpm"},
   "plug_crypto": {:hex, :plug_crypto, "1.0.0", "18e49317d3fa343f24620ed22795ec29d4a5e602d52d1513ccea0b07d8ea7d4d", [:mix], [], "hexpm"},
   "poison": {:hex, :poison, "3.1.0", "d9eb636610e096f86f25d9a46f35a9facac35609a7591b3be3326e99a0484665", [:mix], [], "hexpm"},
```
samfrench: Do we still need poison?

lib/mozart_fetcher/decoder.ex
JoeARO: Yea I forgot to remove the comments, do you think its useful to keep them and move it up?

samfrench: Not if you were going to remove them.

mix.lock
JoeARO: Its a dependency of httpoison I think

ettomatic: mmm

lib/mozart_fetcher/decoder.ex
```diff
@@ -0,0 +1,28 @@
+defmodule MozartFetcher.Decoder do
```
ettomatic: this module needs tests!

ettomatic: ah you probably told me that you were going to add them, ignore in case

JoeARO: Yes yes, will be adding was just to get a quick POC so I could test the branch for performance improvements.

lib/mozart_fetcher/router.ex
```diff
@@ -21,19 +21,36 @@ defmodule MozartFetcher.Router do
...
   end
 
   match _ do
     send_resp(conn, 404, "not found")
   end
 
-  def decode(body) do
-    Poison.decode!(body, as: %{"components" => [%Config{}]})
+  def config(body) do
+    IO.inspect(body)
```
woodyblah: Just a tiny thing, but I don't think we still need this IO.inspect

JoeARO: Thanks

test/decoder_test.exs
```diff
@@ -0,0 +1,45 @@
...
+      }}
+
+      assert Decoder.data_to_struct(json, %Envelope{}) == expected
+    end
+  end
+
+  describe "#list_to_struct_list" do
+    test "when the list of JSON is successfuly decoded it is transformed into a list of structs" do
+      json =  ~s({ "components\": [{"id": "stream-icons", "endpoint": "localhost:8082/success", "must_succeed": true },
+                                  {"id": "stream-icons", "endpoint": "localhost:8082/success", "must_succeed": true }
```
samfrench: Can we set the second one to have different values in case we are transforming the same data twice? It looks to iterate on each one, but just in case.

PR #19, opened 2019-04-08
ID gets passed through to component even upon failure

JoeARO:


ettomatic:
agree with Ryan I'd renamed the ids for documentation purposes

test/component_test.exs
```diff
@@ -7,15 +7,15 @@ defmodule MozartFetcher.ComponentTest do
 
   describe "fetch components" do
     test "it returns the response body when succesfull" do
-      config = %Config{endpoint: "http://localhost:8082/success"}
+      config = %Config{endpoint: "http://localhost:8082/success", id: "123"}
```
woodyblah: Ha I guess these technically are string now 😅

I think as Ettore mentioned in his comment that for documentation purposes it would be better to have the ids in the tests look like "real" ids. eg `"news_top_stories"`, `"news_navigation"`, or `"weather_forecast"` etc. rather than just a number

JoeARO: ohhh I see, sorry that completely slipped my mind as I done it quickly to begin with! Ill sort that now :)

PR #20, opened 2019-04-12
Integration tests

JoeARO:


woodyblah:
Could you also add a test for fetching more than one component?

test/integration_test.exs
```diff
@@ -0,0 +1,251 @@
...
+      assert conn.resp_body == expected_body
+    end
+  end
+
+  describe "when fetching the component receives a server error" do
+    test "it returns a 200, with the failing component status set as 500 and its original id still set" do
+      json_body = ~s({
+        "components": [{
+                        "id": "weather-front-page",
+                        "endpoint": "localhost:8082/non_existent_path",
```
ettomatic: this sounds like mocking a 404 to me

ettomatic: you should use: `get "/non_200_status/:code" do`

JoeARO: sure

test/integration_test.exs
```diff
@@ -0,0 +1,251 @@
...
+      assert conn.resp_body == expected_body
+    end
+  end
+
+  describe "when fetching the component times out" do
+    test "it returns a 200, with the failing component status set as 408 and its original id still set" do
+      json_body = ~s({
+        "components": [{
+                        "id": "weather-front-page",
+                        "endpoint": "localhost:8082/non_200_status/408",
```
woodyblah: There is also `/timeout` which times out for real, but this works too.

test/integration_test.exs
```diff
@@ -0,0 +1,251 @@
...
+              id: "news-front-page",
+              envelope: %{
+                head: [],
+                bodyLast: [],
+                bodyInline: "<DIV id=\"site-container\">"
+              }
+            },
+            %{
+              status: 200,
+              index: 0,
```
samfrench: Don't we expect this to be index 1?

woodyblah: There is a ticket for this. 

https://jira.dev.bbc.co.uk/browse/RESFRAME-2699

PR #21, opened 2019-04-17
Index components as we build component list

JoeARO:


lib/mozart_fetcher/fetcher.ex
```diff
@@ -12,6 +12,7 @@ defmodule MozartFetcher.Fetcher do
   def process(components) do
     ExMetrics.timeframe "function.timing.fetcher.process" do
       components
+      |> Enum.with_index
       |> Enum.map(&Task.async(fn -> Component.fetch(&1) end))
```
samfrench: I thought we might need to specify the index here, but it looks like the integration tests work so it looks like we wouldn't need to.

PR #22, opened 2019-04-17
perf improvements: Add hackney pool & bump exmetrics version (related to RESFRAME-2701)

JoeARO:
Related to: RESFRAME-2701

### DIFF
--- a/lib/mozart_fetcher/application.ex
+++ b/lib/mozart_fetcher/application.ex
@@ -5,6 +5,9 @@ defmodule MozartFetcher.Application do
 
   use Application
 
+  @connection_timeout Application.get_env(:mozart_fetcher, :connection_timeout)
+  @max_connections Application.get_env(:mozart_fetcher, :max_connections)
+
   defp children(env: :test) do
     children(env: :prod) ++
       [
@@ -47,6 +50,11 @@ defmodule MozartFetcher.Application do
   end
 
   defp hackney_setup do
-    [:hackney_pool.child_spec(:origin_pool, timeout: 1000, max_connections: 20000)]
+    [
+      :hackney_pool.child_spec(:origin_pool,
+        timeout: @connection_timeout,
+        max_connections: @max_connections
+      )
+    ]
   end
 end
--- a/lib/mozart_fetcher/application.ex
+++ b/lib/mozart_fetcher/application.ex
@@ -5,8 +5,8 @@ defmodule MozartFetcher.Application do
 
   use Application
 
-  @connection_timeout Application.get_env(:mozart_fetcher, :connection_timeout)
-  @max_connections Application.get_env(:mozart_fetcher, :max_connections)
+  @connection_timeout MozartFetcher.connection_timeout()
+  @max_connections MozartFetcher.max_connections()
 
   defp children(env: :test) do
     children(env: :prod) ++
@@ -40,8 +40,6 @@ defmodule MozartFetcher.Application do
   def start(_type, args) do
     # List all child processes to be supervised
     children = children(args)
-
-    Application.put_env(:mozart_fetcher, :dev_cert_pem, System.get_env("DEV_CERT_PEM"))
 
     # See https://hexdocs.pm/elixir/Supervisor.html
     # for other strategies and supported options
--- a/lib/mozart_fetcher/application.ex
+++ b/lib/mozart_fetcher/application.ex
@@ -43,6 +43,10 @@ defmodule MozartFetcher.Application do
     # See https://hexdocs.pm/elixir/Supervisor.html
     # for other strategies and supported options
     opts = [strategy: :one_for_one, name: MozartFetcher.Supervisor]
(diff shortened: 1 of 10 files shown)


james-bowers:
Load tests, with metrics and hackney improvements:
A test to compare results to performance before these changes:
```
./wrk -t2 -c10 -R50 -d30 <internal-eu-west-2-fetcher>/collect -s ../wrk-script.lua
Running 30s test @ <internal-eu-west-2-fetcher>/collect
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   261.58ms  176.05ms   1.33s    89.39%
    Req/Sec    24.57      2.76    31.00     82.54%
  1495 requests in 30.03s, 731.08MB read
Requests/sec:     49.79
Transfer/sec:     24.35MB
```
and repeated:
```
./wrk -t2 -c10 -R50 -d30 <internal-eu-west-2-fetcher>/collect -s ../wrk-script.lua
Running 30s test @ <internal-eu-west-2-fetcher>/collect
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   141.52ms   21.34ms 183.81ms   73.10%
    Req/Sec    24.75      5.37    30.00     77.50%
  1500 requests in 30.03s, 732.36MB read
Requests/sec:     49.96
Transfer/sec:     24.39MB
```
once more:
```
./wrk -t2 -c10 -R50 -d30 <internal-eu-west-2-fetcher>/collect -s ../wrk-script.lua
Running 30s test @ <internal-eu-west-2-fetcher>/collect
  2 threads and 10 connections
  Thread calibration: mean lat.: 141.087ms, rate sampling interval: 326ms
  Thread calibration: mean lat.: 143.765ms, rate sampling interval: 327ms
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   141.36ms   24.16ms 186.50ms   75.00%
    Req/Sec    24.49      4.92    30.00     77.87%
  1500 requests in 30.03s, 732.36MB read
Requests/sec:     49.95
Transfer/sec:     24.39MB
```

lib/mozart_fetcher/application.ex
```diff
@@ -43,6 +43,10 @@ defmodule MozartFetcher.Application do
     # See https://hexdocs.pm/elixir/Supervisor.html
     # for other strategies and supported options
     opts = [strategy: :one_for_one, name: MozartFetcher.Supervisor]
-    Supervisor.start_link(children, opts)
+    Supervisor.start_link(children ++ hackney_setup(), opts)
+  end
+
+  defp hackney_setup do
+    [:hackney_pool.child_spec(:origin_pool, timeout: 1000, max_connections: 20000)]
```
ettomatic: Probably for fetcher `20,000` is way too much

ettomatic: also the timeout should be hardcoded?

james-bowers: hmm, fetcher makes far more HTTP requests than Ingress per request it receives? I agree it is too many, but perhaps so is Ingress then?

james-bowers: I'll use `Application.get_env(:mozart_fetcher, :timeout)` to get the timeout

ettomatic: agree, way too much, but also in that in this case we would risk overflowing of requests our origin.

ettomatic: Have you seen this? https://github.com/bbc/mozart-fetcher/blob/master/lib/mozart_fetcher/http_client.ex#L7

james-bowers: so are we OK to set the `timeout`, and `recv_timeout` to the same value? This might be misleading as our combined timeout will be the sum of them both?

Useful explanation of the difference: https://github.com/edgurgel/httpoison/issues/211#issuecomment-271968830

ettomatic: `timeout` is the time we use to create a connection - means that the client shouldn't wait that much in case of faults. 
So `timeout` should be much smaller than `recv_timeout`

lib/mozart_fetcher/http_client.ex
```diff
@@ -1,13 +1,19 @@
 defmodule HTTPClient do
   use ExMetrics
 
+  @content_timeout MozartFetcher.content_timeout()
+
   def get(endpoint) do
     ExMetrics.timeframe "function.timing.http_client.get" do
-      cert = Application.get_env(:mozart_fetcher, :dev_cert_pem)
-      timeout = Application.get_env(:mozart_fetcher, :timeout)
+      cert = MozartFetcher.cert()
```
james-bowers: I haven't moved this to an `@cert` attribute, but I think we could. If we did this I believe we would specify the cert file at `mix release` time in the fetcher build repo, and not in the cosmos config. Using the `@cert` might be very slightly more performant?

PR #23, opened 2019-05-01
RESFRAME-2729 - get cert paths for SSL requests

james-bowers:
related but already merged PRs:
https://github.com/bbc/mozart-fetcher-build/pull/14
https://github.com/bbc/mozart-fetcher-build/pull/15

Tested by doing:
`cat james-debug-fetcher-config.json | curl -X POST -d @- http://localhost:8080/collect` with a config that hits internal bbc endpoints and gets a response

### DIFF
--- a/lib/mozart_fetcher.ex
+++ b/lib/mozart_fetcher.ex
@@ -21,8 +21,23 @@ defmodule MozartFetcher do
     Application.get_env(:mozart_fetcher, :max_connections)
   end
 
-  def cert do
-    System.get_env("DEV_CERT_PEM")
+  def request_ssl do
+    case environment() do
+      :prod -> [certfile: cert_path(), cacertfile: ca_cert_path(), keyfile: key_file_path()]
+      _ -> [certfile: System.get_env("DEV_CERT_PEM")]
+    end
+  end
+
+  defp cert_path do
+    System.get_env("cert_file_path")
+  end
+
+  defp ca_cert_path do
+    System.get_env("ca_file_path")
+  end
+
+  defp key_file_path do
+    System.get_env("key_file_path")
   end
 
   def environment do
--- a/lib/mozart_fetcher/http_client.ex
+++ b/lib/mozart_fetcher/http_client.ex
@@ -5,13 +5,11 @@ defmodule HTTPClient do
 
   def get(endpoint) do
     ExMetrics.timeframe "function.timing.http_client.get" do
-      cert = MozartFetcher.cert()
-
       headers = []
 
       options = [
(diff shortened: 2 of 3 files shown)


lib/mozart_fetcher.ex
```diff
@@ -21,8 +21,23 @@ defmodule MozartFetcher do
     Application.get_env(:mozart_fetcher, :max_connections)
   end
 
-  def cert do
-    System.get_env("DEV_CERT_PEM")
+  def request_ssl do
```
ettomatic: maybe a test for this?

james-bowers: done :)

lib/mozart_fetcher.ex
```diff
@@ -21,8 +21,23 @@ defmodule MozartFetcher do
...
+    case environment() do
+      :prod -> [certfile: cert_path(), cacertfile: ca_cert_path(), keyfile: key_file_path()]
+      _ -> [certfile: System.get_env("DEV_CERT_PEM")]
+    end
+  end
+
+  defp cert_path do
+    System.get_env("cert_file_path")
```
samfrench: Should we use the same convention for our environment variables, by uppercasing them all?

lib/mozart_fetcher.ex
```diff
@@ -21,8 +21,23 @@ defmodule MozartFetcher do
...
+  end
+
+  defp ca_cert_path do
+    System.get_env("ca_file_path")
+  end
+
+  defp key_file_path do
+    System.get_env("key_file_path")
```
samfrench: Are we keeping the same naming we use in requester, such as key_location?

james-bowers: So I've intentionally not used requester's naming as I didn't think the naming for that was very good, as it switched between ending `_file` and `_location`. Now they all end in the same format, `_file_path`

lib/mozart_fetcher.ex
james-bowers: can do, we didn't for requester, and these environment variables are initially set in the cosmos config, and the convention there is for the names to be lower case.

woodyblah: Maybe we should use capital letters for the cosmos config keys? https://cosmos.tools.bbc.co.uk/services/ingress/test/configuration we have INSTANCE_ROLE_NAME but also `td_ingest_bucket` which is required to be lower case by the td agent :(

james-bowers: yeah that'd be the way to do it, but I'm not fussed either, I think I'll merge this

PR #25, opened 2019-05-07
Support ares/raw json data responses

woodyblah:
To match how Mozart-Requester works this PR implements handling of requests for `"format": "ares"` components.

Ares components are not returned from the origin in the BBC Envelope format, and could be in any data structure, and Requester currently puts these generic data objects into the `data:` key of the component response rather than the `envelope:` key used by normal envelope components.

The work done to modify fetcher to handle this is here: 
https://github.com/bbc/mozart-requester/pull/157

I believe that work could be done to abstract this further, and have fetcher not care at all about envelope/ares format and just return whatever it gets from the origins, but change would need to also be made to the ares renderer in Assembler.

### DIFF
--- a/lib/mozart_fetcher/component.ex
+++ b/lib/mozart_fetcher/component.ex
@@ -1,8 +1,5 @@
 defmodule MozartFetcher.Component do
-  alias MozartFetcher.{Component, Config, Envelope, LocalCache}
-
-  @derive [Jason.Encoder]
-  defstruct [:index, :id, :status, :envelope]
+  alias MozartFetcher.{Components.EnvelopeComponent, Config, Envelope, LocalCache}
 
   def fetch({config = %Config{}, component_index}) do
     process(component_index, config, get(config))
@@ -11,14 +8,14 @@ defmodule MozartFetcher.Component do
   defp process(component_index, config, {:ok, %HTTPoison.Response{status_code: 200, body: body}}) do
     ExMetrics.increment("success.component.process")
     ExMetrics.increment("success.component.process.#{config.id}.200")
-    %Component{index: component_index, id: config.id, status: 200, envelope: Envelope.build(body)}
+    %EnvelopeComponent{index: component_index, id: config.id, status: 200, envelope: Envelope.build(body)}
   end
 
   defp process(component_index, config, {:ok, %HTTPoison.Response{status_code: status_code, body: body}}) do
     ExMetrics.increment("error.component.process")
     ExMetrics.increment("error.component.process.#{config.id}.#{status_code}")
     Stump.log(:error, %{message: "Non-200 response. Got status:#{status_code} for component #{config.id}"})
-    %Component{index: component_index, id: config.id, status: status_code, envelope: %Envelope{}}
+    %EnvelopeComponent{index: component_index, id: config.id, status: status_code, envelope: %Envelope{}}
   end
 
   defp process(component_index, config, {:error, %HTTPoison.Error{reason: reason}}) do
@@ -33,10 +30,10 @@ defmodule MozartFetcher.Component do
   end
 
   defp failed_component(component_index, :timeout, id) do
-    %Component{index: component_index, id: id, status: 408, envelope: %Envelope{}}
+    %EnvelopeComponent{index: component_index, id: id, status: 408, envelope: %Envelope{}}
   end
 
   defp failed_component(component_index, _, id) do
-    %Component{index: component_index, id: id, status: 500, envelope: %Envelope{}}
+    %EnvelopeComponent{index: component_index, id: id, status: 500, envelope: %Envelope{}}
   end
 end
--- a/lib/mozart_fetcher/component.ex
+++ b/lib/mozart_fetcher/component.ex
@@ -68,12 +68,16 @@ defmodule MozartFetcher.Component do
 
   defp metric(id, status) when is_integer(status) do
     ExMetrics.increment("error.component.process")
+    ExMetrics.increment("error.component.process.#{id}")
     ExMetrics.increment("error.component.process.#{id}.#{status}")
+    ExMetrics.increment("error.component.process.#{status}")
     Stump.log(:error, %{message: "Non-200 response. Got status:#{status} for component #{id}"})
   end
 
   defp metric(id, reason) do
     ExMetrics.increment("error.component.process")
+    ExMetrics.increment("error.component.process.#{id}")
+    ExMetrics.increment("error.component.process.#{reason}")
     ExMetrics.increment("error.component.process.#{id}.#{reason}")
     Stump.log(:error, %{message: "Failed to process HTTP request, reason: #{reason}"})
   end
--- a/lib/mozart_fetcher/component.ex
+++ b/lib/mozart_fetcher/component.ex
@@ -1,21 +1,35 @@
 defmodule MozartFetcher.Component do
-  alias MozartFetcher.{Components.EnvelopeComponent, Config, Envelope, LocalCache}
+  alias MozartFetcher.{Components, Components.Ares, Config, Envelope, LocalCache}
 
   def fetch({config = %Config{}, component_index}) do
     process(component_index, config, get(config))
   end
 
+  #todo refactor and handle decoding errors, add more tests for invalid json.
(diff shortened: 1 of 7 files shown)


lib/mozart_fetcher/components/ares.ex
```diff
@@ -0,0 +1,7 @@
+defmodule MozartFetcher.Components.Ares do
+  alias MozartFetcher.Components.Ares
+  @moduledoc false
+  @derive [Jason.Encoder]
+  defstruct [:index, :id, :status, :data]
+  @type t :: %Ares{index: non_neg_integer(), id: String.t(), status: non_neg_integer(), data: Map.t()}
```
james-bowers:
```suggestion
  @type t :: %__MODULE__{index: non_neg_integer(), id: String.t(), status: non_neg_integer(), data: Map.t()}
```
I _think_ you can do that, then there is no need for the alias on line 2.

lib/mozart_fetcher/components/envelope.ex
```diff
@@ -0,0 +1,7 @@
+defmodule MozartFetcher.Components.Envelope do
```
james-bowers: not sure how I feel about having `MozartFetcher.Envelope`, and `MozartFetcher.Components.Envelope` modules, maybe its completely fine, just starting the conversation :)

lib/mozart_fetcher/component.ex
```diff
@@ -1,24 +1,35 @@
...
+
+  defp process(component_index, config = %Config{format: "ares"}, {:ok, %HTTPoison.Response{status_code: status_code, body: _body}}) do
+    ExMetrics.increment("error.component.process")
+    ExMetrics.increment("error.component.process.#{config.id}.#{status_code}")
+    Stump.log(:error, %{message: "Non-200 response. Got status:#{status_code} for component #{config.id}"})
+    %Components.Ares{index: component_index, id: config.id, status: status_code, data: %{}}
   end
 
-  defp process(component_index, config, {:ok, %HTTPoison.Response{status_code: status_code, body: body}}) do
+  defp process(component_index, config, {:ok, %HTTPoison.Response{status_code: status_code, body: _body}}) do
```
james-bowers:
```suggestion
  defp process(component_index, config, {:ok, %HTTPoison.Response{status_code: status_code}}) do
```

lib/mozart_fetcher/components/ares.ex
```diff
@@ -0,0 +1,7 @@
+defmodule MozartFetcher.Components.Ares do
+  alias MozartFetcher.Components.Ares
+  @moduledoc false
+  @derive [Jason.Encoder]
```
james-bowers:
```suggestion
```
I don't think we need this anymore, as Jason doesn't support the option `as` like Poison did. https://github.com/michalmuskala/jason#differences-to-poison

lib/mozart_fetcher/components/envelope.ex
woodyblah: Yeah this is one of things I wanted to think about changing. Components.EnvelopeComponent felt unnecessary but calling them both Envelope is also confusing..

lib/mozart_fetcher/components/ares.ex
woodyblah: The link suggests that you DO have to derive the encoder if you want to encode structs so I'm a bit confused 🤔 

Good point about potentially not needing them to be structs at all though, I like it for the type safety but perhaps it's adding unnecessary complications

james-bowers: whoops yeah sorry, I got muddled between encoding and decoding... :/

lib/mozart_fetcher/component.ex
```diff
@@ -1,24 +1,35 @@
...
+  end
+
   defp process(component_index, config, {:ok, %HTTPoison.Response{status_code: 200, body: body}}) do
     ExMetrics.increment("success.component.process")
     ExMetrics.increment("success.component.process.#{config.id}.200")
-    %Component{index: component_index, id: config.id, status: 200, envelope: Envelope.build(body)}
+    %Components.Envelope{index: component_index, id: config.id, status: 200, envelope: Envelope.build(body)}
+  end
+
+  defp process(component_index, config = %Config{format: "ares"}, {:ok, %HTTPoison.Response{status_code: status_code, body: _body}}) do
```
james-bowers:
```suggestion
  defp process(component_index, config = %Config{format: "ares"}, {:ok, %HTTPoison.Response{status_code: status_code}}) do
```

samfrench:
Why do we need to do something for Ares/JSON in Fetcher? I thought Requester just sent back what it received, so wondering what the difference is.

Can we add a description to say why it is not supported.

lib/mozart_fetcher/component.ex
```diff
@@ -1,42 +1,80 @@
...
+    %{index: component_index, id: config.id, status: status_code, data: %{}}
+  end
+
+  defp process(component_index, config, {:ok, %HTTPoison.Response{status_code: status_code}}) do
+    metric(config.id, status_code)
+    %{index: component_index, id: config.id, status: status_code, envelope: %Envelope{}}
+  end
+
+  defp process(component_index, config = %Config{format: "ares"}, {:error, %HTTPoison.Error{reason: reason}}) do
+    metric(config.id, reason)
```
samfrench: What can reason be? If it's a message, this could be long in the metric name.

lib/mozart_fetcher/component.ex
```diff
@@ -1,42 +1,80 @@
...
   end
 
-  defp failed_component(component_index, :timeout, id) do
-    %Component{index: component_index, id: id, status: 408, envelope: %Envelope{}}
+  defp failed_component(component_index, :timeout, id, :envelope) do
+    %{index: component_index, id: id, status: 408, envelope: %Envelope{}}
+  end
+
+  defp failed_component(component_index, _, id, :envelope) do
+    %{index: component_index, id: id, status: 500, envelope: %Envelope{}}
```
samfrench: I wonder if at a later date we should just return envelopes as the data key and we use the upstream service to check for envelope keys or data. If we trust fetcher more, we won't necessarily need to do a lot of validation.

lib/mozart_fetcher/component.ex
woodyblah: An atom with a (usually one word) reason why the connection failed. I couldn't find a full list but the doccumentation mentions `:closed`, `:econnrefused`, and we see `:timeout` in the tests and in prod

lib/mozart_fetcher/component.ex
woodyblah: I agree, this extra `data:` key only really exists because I found Requester hard to change

woodyblah:
@samfrench 
> I thought Requester just sent back what it received

This is not  the case unfortunately, it marshals the data it receives int an Envelope struct (just like we do here in the Elixir version) which unfortunately means Ares data needs handling differently as it does not marshal into the Envelope struct properly. I am sure that fetcher/requester could be data agnostic, and not care about ares/envelope/whatever data types but this would be a future improvement, this is just a fix to get the fetcher behaving the same as requester

lib/mozart_fetcher/component.ex
samfrench: I wonder if we should either hardcode "error" or use a status. We have "component-200" for successful ones, so we could have "component-404" or "component-500" when they are non-200 responses.

Otherwise to see how many errors of a particular component, we'd have a lot to add in Grafana.

woodyblah: We already have status code ones for non 200 responses. These errors happen when we don't even receive a status code. I believe we will still want to see metrics for these.

I think the solution is not fewer, and more generic metrics, but rather just adding more metrics.
We should keep the `error.component.process.#{id}.#{status}"` and `error.component.process.#{id}.#{reason}"` but also include a `error.component.process.#{id}`, `error.component.process.#{reason}"`, and `error.component.process.#{status}"` metric so we can graph different situations better

lib/mozart_fetcher/components/envelope.ex
woodyblah: Solved by not using structs for Components. Good suggestion! :)

lib/mozart_fetcher/component.ex
woodyblah: I've just pushed a commit adding these new alarms so we can create all the graphs and alarms as we need

PR #26, opened 2019-05-08
Get cache ttl value from system environment var

woodyblah:


lib/mozart_fetcher/application.ex
```diff
@@ -55,4 +55,12 @@ defmodule MozartFetcher.Application do
       )
     ]
   end
+
+  defp to_ttl(nil), do: 10
+  defp to_ttl(ttl) when is_integer(ttl), do: ttl
+  defp to_ttl(ttl) do
+    String.to_integer(ttl)
+  rescue
+    ArgumentError -> 10
```
ettomatic: I'd log this occurrence maybe...

lib/mozart_fetcher/application.ex
```diff
@@ -32,7 +32,7 @@ defmodule MozartFetcher.Application do
        [
          name: :fetcher_cache,
          ttl_check_interval: :timer.seconds(1),
-         global_ttl: :timer.seconds(30)
+         global_ttl: :timer.seconds(to_ttl(System.get_env("local_cache_ttl")))
```
samfrench: I wonder if we should pick either upper or lower case. Not for this PR, but I think we use a few different casing conventions.

samfrench: This will be seconds and not milliseconds. The default timeout was 3000, but I don't see that in the build repo.

woodyblah: Correct, seconds not millisecond.

Also t seems Fetcher uses lower case across the board, the one exception is DEV_CERT_PEM for the cert path on local development

PR #28, opened 2019-05-14
Updates Fetcher version.

ettomatic:


mix.exs
```diff
@@ -4,8 +4,8 @@ defmodule MozartFetcher.MixProject do
   def project do
     [
       app: :mozart_fetcher,
-      version: "0.1.0",
-      elixir: "~> 1.6",
+      version: "1.0.0",
+      elixir: "~> 1.7",
```
samfrench: Why not 1.8?

ettomatic: The Docker image we are using is at 1.7.x
https://github.com/qixxit/docker/blob/master/elixir-centos/Dockerfile

PR #30, opened 2019-05-14
Logs component endpoint URL.

ettomatic:


lib/mozart_fetcher/component.ex
```diff
@@ -61,24 +61,24 @@ defmodule MozartFetcher.Component do
...
   end
 
-  defp metric(id, status) when is_integer(status) do
+  defp metric(id, endpoint, status) when is_integer(status) do
     ExMetrics.increment("error.component.process")
     ExMetrics.increment("error.component.process.#{id}")
     ExMetrics.increment("error.component.process.#{id}.#{status}")
     ExMetrics.increment("error.component.process.#{status}")
-    Stump.log(:error, %{message: "Non-200 response. Got status:#{status} for component #{id}"})
+    Stump.log(:error, %{message: "Non-200 response. Got status:#{status} for component #{id}, endpoint: #{endpoint}"})
```
samfrench: Does this convert to JSON easily? Could we have this format as it will make it easier to search if they're separate fields:
```rb
{
  message: "Non-200 response",
  status: status,
  component: id,
  endpoint: endpoint
}
```

lib/mozart_fetcher/component.ex
```diff
@@ -61,24 +61,27 @@ defmodule MozartFetcher.Component do
...
+  defp metric(id, endpoint, status) when is_integer(status) do
     ExMetrics.increment("error.component.process")
     ExMetrics.increment("error.component.process.#{id}")
     ExMetrics.increment("error.component.process.#{id}.#{status}")
     ExMetrics.increment("error.component.process.#{status}")
-    Stump.log(:error, %{message: "Non-200 response. Got status:#{status} for component #{id}"})
+    Stump.log(:error, %{message: "Non-200 response",
+                        status: status,
+                        component: id,
+                        endpoint: endpoint})
```
james-bowers: just double checking that we never fetch from endpoints that have the `?api_key=` query string value? I know we do for context calls, so I'm just checking that isn't the same here so we don't log api keys

ettomatic: Good point! I can't think at anything but could happen in the future....

lib/mozart_fetcher/component.ex
james-bowers: this is marked as outdated, but `status` and `component` still aren't their own keys, they are still interpolated into the message. I think having them as Sam's suggestion would be neater

lib/mozart_fetcher/component.ex
```diff
@@ -61,24 +61,27 @@ defmodule MozartFetcher.Component do
...
   end
 
-  defp metric(id, reason) do
+  defp metric(id, endpoint, reason) do
     ExMetrics.increment("error.component.process")
     ExMetrics.increment("error.component.process.#{id}")
     ExMetrics.increment("error.component.process.#{reason}")
     ExMetrics.increment("error.component.process.#{id}.#{reason}")
-    Stump.log(:error, %{message: "Failed to process HTTP request, reason: #{reason}, id: #{id}"})
+    Stump.log(:error, %{message: "Failed to process HTTP request, reason: #{reason}, id: #{id}, endpoint: #{endpoint}"})
```
samfrench: Can we change the format of this to match the above with the keys in the log.

lib/mozart_fetcher/component.ex
samfrench: At the moment we don't have any api keys for components. We need to log query strings as we currently do this in Requester and this shows the component timeout or query parameters we pass to brokers. If we don't have this, it would be hard to debug any request errors.

ettomatic: ok I was already working on it but makes sense, let's keep the qs for now in the logs

PR #33, opened 2019-05-17
Allow retry for closed response

samfrench:
This performs the request then matches for a closed connection and then performs the request again if this is the case.

It seems as connections are reused, some requests might be happening while the connection is closed causing an error with closed being the reason.

### DIFF
--- a/lib/mozart_fetcher/http_client.ex
+++ b/lib/mozart_fetcher/http_client.ex
@@ -13,18 +13,26 @@ defmodule HTTPClient do
         hackney: [pool: :origin_pool]
       ]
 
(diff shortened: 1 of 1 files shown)


lib/mozart_fetcher/http_client.ex
```diff
@@ -13,18 +13,26 @@ defmodule HTTPClient do
         hackney: [pool: :origin_pool]
       ]
 
-      valid_response?(HTTPoison.get(sanitise(endpoint), headers, options))
+      case make_request(sanitise(endpoint), headers, options) do
+        {:error, %HTTPoison.Error{reason: "closed"}} ->
```
ettomatic: What happened to the approach you showed me? It looked nice :)

PR #34, opened 2019-05-19
Second attempt at retrying failed requests

samfrench:
The http_client was split into making the request and retrying failed connections and any response parsing and logging was moved into a component_fetcher. Module name can be discussed.

Both of these were tested by adding a client parameter to allow for a mock client to be passed in. The issue with this is returning back different data, unsuccessful and then successful for the same requests. I had a few ideas around this, but best to keep this simple. This was in a different direction to using a mocking library as a functional language should allow for dependencies to be passed in.

Any ideas to simplify this are welcome.

### DIFF
--- a/lib/mozart_fetcher/http_client.ex
+++ b/lib/mozart_fetcher/http_client.ex
@@ -3,7 +3,7 @@ defmodule HTTPClient do
 
   @content_timeout MozartFetcher.content_timeout()
 
-  def get(endpoint) do
+  def get(endpoint, client \\ client()) do
     ExMetrics.timeframe "function.timing.http_client.get" do
       headers = []
 
@@ -13,26 +13,36 @@ defmodule HTTPClient do
         hackney: [pool: :origin_pool]
       ]
 
-      case make_request(sanitise(endpoint), headers, options) do
-        {:error, %HTTPoison.Error{reason: "closed"}} ->
-          handle_response(make_request(sanitise(endpoint), headers, options))
-        {k, response} -> handle_response({k, response})
+      case log_errors_and_return(make_request(sanitise(endpoint), headers, options, client)) do
+        {:error, %HTTPoison.Error{reason: :closed}} ->
+          make_request(sanitise(endpoint), headers, options, client)
+
+        {:error, %HTTPoison.Error{reason: :timeout}} ->
+          make_request(sanitise(endpoint), headers, options, client)
+
+        {k, resp} ->
+          {k, resp}
       end
+      |> log_errors_and_return()
     end
   end
-
-  defp handle_response(response = {:error, %HTTPoison.Error{reason: reason}}) do
-    Stump.log(:error, %{message: "HTTPoison Error", reason: reason})
-    response
-  end
-
-  defp handle_response(response), do: response
 
   defp sanitise(endpoint) do
     String.replace(endpoint, " ", "%20")
   end
 
-  defp make_request(endpoint, headers, options) do
-    HTTPoison.get(endpoint, headers, options)
+  defp make_request(endpoint, headers, options, client) do
+    client.get(endpoint, headers, options)
+  end
+
+  defp log_errors_and_return(response = {:error, %HTTPoison.Error{reason: reason}}) do
+    Stump.log(:error, %{message: "HTTPoison Error", reason: reason})
+    response
+  end
+
+  defp log_errors_and_return(response = {:ok, _}), do: response
+
+  defp client() do
+    HTTPoison
   end
 end
--- a/test/http_client_test.exs
+++ b/test/http_client_test.exs
@@ -26,5 +26,122 @@ defmodule HTTPClientTest do
       {:ok, resp} = HTTPClient.get("http://localhost:8082/foo bar?baz= true")
       assert resp.request_url == "http://localhost:8082/foo%20bar?baz=%20true"
     end
+
+    test "makes only one request on success" do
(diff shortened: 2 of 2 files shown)


lib/mozart_fetcher/http_client.ex
```diff
@@ -13,26 +13,28 @@ defmodule HTTPClient do
         hackney: [pool: :origin_pool]
       ]
 
-      case make_request(sanitise(endpoint), headers, options) do
-        {:error, %HTTPoison.Error{reason: "closed"}} ->
-          handle_response(make_request(sanitise(endpoint), headers, options))
-        {k, response} -> handle_response({k, response})
+      case make_request(sanitise(endpoint), headers, options, client) do
```
samfrench: I tried to turn this into this, but the scope of the resp inside the if statement was not the same:
```ex
  if retry_if_unsuccessful?({k, resp}) do
    {k, resp} = make_request(sanitise(endpoint), headers, options, client)
  end

  {k, resp}
end

...

defp retry_if_unsuccessful?({:error, %HTTPoison.Error{reason: :closed}}), do: true

defp retry_if_unsuccessful?({:error, %HTTPoison.Error{reason: :timeout}}), do: true

defp retry_if_unsuccessful?(_), do: false
```

lib/mozart_fetcher/component.ex
```diff
@@ -42,7 +42,7 @@ defmodule MozartFetcher.Component do
   end
 
   defp get(config) do
-    LocalCache.get_or_store(config.endpoint, fn -> HTTPClient.get(config.endpoint) end)
+    LocalCache.get_or_store(config.endpoint, fn -> ComponentFetcher.fetch(config.endpoint) end)
```
samfrench: Naming for this can be decided/changed at later date. This could be ComponentParser or ResponseHandler.

@ettomatic may have some thoughts on this.

lib/mozart_fetcher/component_fetcher.ex
```diff
@@ -0,0 +1,18 @@
+defmodule MozartFetcher.ComponentFetcher do
```
samfrench: This could go back into the http client as this doesn't do much at the moment.

test/component_fetcher_test.exs
```diff
@@ -0,0 +1,38 @@
...
+                  headers: [],
+                  request: nil,
+                  request_url: nil,
+                  status_code: nil
+                }}
+    end
+  end
+
+  test "unsuccessful response" do
+    defmodule MockClient do
```
samfrench: When defining this mock for another test I get a warning about redefining. I could rename this, but maybe there is another way to silence the warnings.

lib/mozart_fetcher/component_fetcher.ex
```diff
@@ -0,0 +1,18 @@
...
+  use ExMetrics
```
james-bowers:
```suggestion
```

james-bowers:
I think we should look at using `Mox` for the mocking. https://hexdocs.pm/mox/Mox.html

test/component_fetcher_test.exs
```diff
@@ -0,0 +1,38 @@
+defmodule MozartFetcher.ComponentFetcherTest do
+  use ExUnit.Case
+
+  alias MozartFetcher.{ComponentFetcher}
+
+  describe "fetch" do
+    test "successful response" do
+      defmodule MockClient do
+        def get(_) do
```
james-bowers: Instead of `_`, we could match on the exact string we expect to be sent to the `get` function

ettomatic:
I think we need to do a proper discussion about Mox and friends :)

Mocking is always hard and We should review what we are doing and agree on a shared approach. This shouldn't block this PR from being merged

lib/mozart_fetcher/http_client.ex
```diff
@@ -13,26 +13,36 @@ defmodule HTTPClient do
...
 
-      case make_request(sanitise(endpoint), headers, options) do
-        {:error, %HTTPoison.Error{reason: "closed"}} ->
-          handle_response(make_request(sanitise(endpoint), headers, options))
-        {k, response} -> handle_response({k, response})
+      case log_errors_and_return(make_request(sanitise(endpoint), headers, options, client)) do
+        {:error, %HTTPoison.Error{reason: :closed}} ->
+          make_request(sanitise(endpoint), headers, options, client)
+
+        {:error, %HTTPoison.Error{reason: :timeout}} ->
```
james-bowers: if upstream starts failing under the high load, do we want to retry?

ettomatic: once yes

woodyblah: I think it's irrelevant, assembler waits a bit longer than 3 seconds then gives up. If we timeout and retry assembler probably wont be listening.

lib/mozart_fetcher/http_client.ex
```diff
@@ -13,26 +13,36 @@ defmodule HTTPClient do
...
+          make_request(sanitise(endpoint), headers, options, client)
+
+        {k, resp} ->
+          {k, resp}
       end
+      |> log_errors_and_return()
```
woodyblah: Does this mean we call `log_errors_and_return()` twice? It's also done on line 16

samfrench: Yes. I have just removed one of them in the later PR.

PR #35, opened 2019-05-20
Refine logging and remove timeout retry

samfrench:
Remove the first metric and logging for error as we retry. If the retry was in the library we wouldn't be logging this anyway.

Removes the timeout matcher so we only retry on connection closed.

### DIFF
--- a/lib/mozart_fetcher/component.ex
+++ b/lib/mozart_fetcher/component.ex
@@ -67,7 +67,6 @@ defmodule MozartFetcher.Component do
   end
 
   defp metric(id, endpoint, status) when is_integer(status) do
-    ExMetrics.increment("error.component.process")
     ExMetrics.increment("error.component.process.#{id}")
     ExMetrics.increment("error.component.process.#{id}.#{status}")
     ExMetrics.increment("error.component.process.#{status}")
--- a/lib/mozart_fetcher/http_client.ex
+++ b/lib/mozart_fetcher/http_client.ex
@@ -15,6 +15,7 @@ defmodule HTTPClient do
 
       case make_request(sanitise(endpoint), headers, options, client) do
         {:error, %HTTPoison.Error{reason: :closed}} ->
+          ExMetrics.increment("http.component.retry")
           make_request(sanitise(endpoint), headers, options, client)
 
         {k, resp} ->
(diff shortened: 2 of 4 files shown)


lib/mozart_fetcher/http_client.ex
```diff
@@ -13,13 +13,10 @@ defmodule HTTPClient do
         hackney: [pool: :origin_pool]
       ]
 
-      case log_errors_and_return(make_request(sanitise(endpoint), headers, options, client)) do
+      case make_request(sanitise(endpoint), headers, options, client) do
         {:error, %HTTPoison.Error{reason: :closed}} ->
           make_request(sanitise(endpoint), headers, options, client)
```
james-bowers: perhaps we could have a metric specifically for retries?

lib/mozart_fetcher/component.ex
```diff
@@ -67,7 +67,6 @@ defmodule MozartFetcher.Component do
   end
 
   defp metric(id, endpoint, status) when is_integer(status) do
-    ExMetrics.increment("error.component.process")
     ExMetrics.increment("error.component.process.#{id}")
```
james-bowers:
```suggestion
    ExMetrics.increment("error.component.process.#{id}")
    ExMetrics.increment("http.component.retry")
```

lib/mozart_fetcher/component.ex
```diff
@@ -67,7 +67,6 @@ defmodule MozartFetcher.Component do
   end
 
   defp metric(id, endpoint, status) when is_integer(status) do
-    ExMetrics.increment("error.component.process")
```
james-bowers: this will need a PR to remove this from grafana

samfrench: This still exists for the other metric method where we metric for a reason. It's the same one, but a bit confusing it does both.

lib/mozart_fetcher/component.ex
samfrench: I added it to the http_client.ex file. Isn't the component the wrong place as we won't know if it's a retry or not?

lib/mozart_fetcher/component.ex
james-bowers: sorry i didn't spot this metric was still being used

lib/mozart_fetcher/component.ex
james-bowers: sorry, for some reason i thought this was in the `config/metrics.exs` file

PR #36, opened 2019-05-20
Cache successul responses

ettomatic:
Uses and tests the cache layer on all environment (we could consider to remove it from dev/localdev in future if make sense...)

Only caches if http response is 200.

### DIFF
--- a/lib/mozart_fetcher/local_cache.ex
+++ b/lib/mozart_fetcher/local_cache.ex
@@ -1,8 +1,8 @@
 defmodule MozartFetcher.LocalCache do
   def get_or_store(id, f) do
-    case MozartFetcher.environment() do
-      :prod -> ConCache.get_or_store(:fetcher_cache, id, f)
-      _ -> f.()
+    case f.() do
+      {:ok, resp} -> ConCache.get_or_store(:fetcher_cache, id, fn() -> {:ok, resp} end)
+      {:error, resp} -> {:error, resp}
     end
   end
 end
--- a/lib/mozart_fetcher/local_cache.ex
+++ b/lib/mozart_fetcher/local_cache.ex
@@ -1,8 +1,22 @@
 defmodule MozartFetcher.LocalCache do
-  def get_or_store(id, f) do
-    case f.() do
-      {:ok, resp} -> ConCache.get_or_store(:fetcher_cache, id, fn() -> {:ok, resp} end)
-      {:error, resp} -> {:error, resp}
+  def get_or_store(id, fun) do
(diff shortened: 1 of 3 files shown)


lib/mozart_fetcher/local_cache.ex
```diff
@@ -1,8 +1,22 @@
...
-      :prod -> ConCache.get_or_store(:fetcher_cache, id, f)
-      _ -> f.()
+  def get_or_store(id, fun) do
+    ConCache.get(:fetcher_cache, id)
+    |> cond_fetch(id, fun)
+  end
+
+  defp cond_fetch(nil, id, fun) do
+    case fun.() do
+      resp = {:ok, %HTTPoison.Response{status_code: 200}} -> store(id, resp)
```
james-bowers: did you mean this?
```suggestion
      resp -> {:ok, %HTTPoison.Response{status_code: 200}} -> store(id, resp)
```

lib/mozart_fetcher/local_cache.ex
```diff
@@ -1,8 +1,22 @@
...
+      other -> other
     end
   end
+
+  defp cond_fetch(value, _id, _fun) do
+    value
+  end
+
+  defp store(id, resp) do
+    ConCache.put(:fetcher_cache, id, resp)
```
ettomatic: unfortunately `put` returns `:ok`

lib/mozart_fetcher/local_cache.ex
```diff
@@ -1,8 +1,22 @@
 defmodule MozartFetcher.LocalCache do
-  def get_or_store(id, f) do
-    case MozartFetcher.environment() do
```
woodyblah: Does this mean we now store stuff in the cache when running the unit tests? The services do not restart between tests automatically, could the cache cause issues?

ettomatic: discussed face 2 face, but shortly, given that we use caching in a single layer coudl make it seems to make more sense to cache in all environment, al least for now.

ettomatic: we'll regret and fix thsi when issues with testing downstream services will arise

lib/mozart_fetcher/local_cache.ex
woodyblah: I think this means there is a chance it could return `:error` so I think the _perfect_ approach would be to match on that potential eventuality (maybe just log and metric then return `resp` anyway) but I think this is unlikely and it's fine as is

PR #39, opened 2019-05-21
Query string timeout

ettomatic:
Allows parsing timeout from component query strings:
```
http://api.site.com/data/page-styles/preload/true/version/2.0.9?timeout=5
```

### DIFF
--- a/lib/mozart_fetcher/timeout_parser.ex
+++ b/lib/mozart_fetcher/timeout_parser.ex
@@ -10,9 +10,9 @@ defmodule MozartFetcher.TimeoutParser do
     @default_timeout
   end
 
-  defp query(qs) do
-    q = URI.decode_query(qs)
-    fetch(q["timeout"])
+  defp query(q) do
+    qs = URI.decode_query(q)
+    fetch(qs["timeout"])
   end
 
   defp fetch(nil) do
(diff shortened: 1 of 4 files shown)


lib/mozart_fetcher/timeout_parser.ex
```diff
@@ -0,0 +1,37 @@
...
+
+  defp fetch(nil) do
+    @default_timeout
+  end
+
+  defp fetch("0") do
+    @default_timeout
+  end
+
+  defp fetch(qs_timeout) do
```
james-bowers: just checking this is always guaranteed to be a string? Because `Integer.parse(<integer>)` will throw an error

ettomatic: it _should_ as `URI.parse` always produces qs values as string

samfrench: I have wrapped that earlier in `parse_timeout` which could handle an error. If so, this will just be the default timeout.

`URI.parse` could produce a nil for the qs value, but other than that it will be a string. If we use Integer.parse on a string such as "abc" then this will just be replaced with the default timeout.

There are tests for all of these situations.

lib/mozart_fetcher/timeout_parser.ex
```diff
@@ -0,0 +1,37 @@
...
+    parse_timeout(Integer.parse(qs_timeout))
```
woodyblah: I like `String.to_integer` more than `Integer.parse` as it simply returns the integer value, not the tuple with the remainder, this can make it easier to handle when piping (which might have looked nicer here than nesting the method calls) as you do need to destructure it but it's only a minor difference. On the downside `Integer.parse` can handle strings with decimals like "2.5" but `String.to_integer` would fail

PR #40, opened 2019-05-21
Add component max timeout to task await

samfrench:


lib/mozart_fetcher/fetcher.ex
```diff
@@ -12,14 +12,21 @@ defmodule MozartFetcher.Fetcher do
   def process(components) do
     ExMetrics.timeframe "function.timing.fetcher.process" do
       components
-      |> Enum.with_index
+      |> Enum.with_index()
       |> Enum.map(&Task.async(fn -> Component.fetch(&1) end))
-      |> Enum.map(&Task.await/1)
+      |> Enum.map(&Task.await(&1, max_timeout(components)))
```
ettomatic: aren't we `await`ing now for every single item in the component list?

lib/mozart_fetcher/fetcher.ex
```diff
@@ -12,14 +12,21 @@ defmodule MozartFetcher.Fetcher do
...
       |> decorate_response
       |> Jason.encode!()
     end
   end
 
+  defp max_timeout(components) do
```
ettomatic: wondering if this could go in the Timeout Parser module

ettomatic: TimeoutParser.max(components.endpoint)

ettomatic: so it could make this bit of code easily testable

lib/mozart_fetcher/fetcher.ex
```diff
@@ -12,14 +12,21 @@ defmodule MozartFetcher.Fetcher do
...
+    components
+    |> Enum.map(&Map.get(&1, :endpoint))
```
james-bowers: instead of 2 maps, how about this reduce?
`Enum.reduce(components, 0, &Enum.max([TimeoutParser.parse(&1.endpoint), &2]))`
we can expand it to make it more readable...

lib/mozart_fetcher/fetcher.ex
ettomatic: in any case I'd pass the componens subesect as it would make things a bit more generic

lib/mozart_fetcher/fetcher.ex
ettomatic: is it really faster with 50 items?

ettomatic: we should *benchmark it* ! :)

lib/mozart_fetcher/fetcher.ex
james-bowers: 👍 I agree

lib/mozart_fetcher/fetcher.ex
james-bowers: yes we should, but surely it is? Instead of iterating through the array 150 times, we do 52 times?

lib/mozart_fetcher/fetcher.ex
samfrench: I have moved the maximum timeout logic to the timeout parser.

lib/mozart_fetcher/timeout_parser.ex
```diff
@@ -6,6 +6,13 @@ defmodule MozartFetcher.TimeoutParser do
     query(uri.query)
   end
 
+  def max_timeout(components) do
+    components
```
james-bowers: Same discussion:
https://github.com/bbc/mozart-fetcher/pull/40#discussion_r286063590

lib/mozart_fetcher/fetcher.ex
ettomatic: our array shouldn't be that big, and sometimes the cost of instantiating new intermediate object can be counterproductive. This is not to say we shouldn't try, instead you're probably right! Just suggesting to be data driven

samfrench: Didn't the data show us an issue!

lib/mozart_fetcher/timeout_parser.ex
samfrench: See PR for this:
https://github.com/bbc/mozart-fetcher/pull/42

PR #42, opened 2019-05-22
Add more time for task await and rename max_timeout to max

samfrench:
Fetcher seems to return one or two 500s every few minutes. This PR allows the task waiting time to be slightly higher than the component fetching time to hopefully reduce the 500s. These 500s only started happening when changing the task await timeout to be the same as the component timeout. Previously this had a default of 5000.

### DIFF
--- a/lib/mozart_fetcher/fetcher.ex
+++ b/lib/mozart_fetcher/fetcher.ex
@@ -11,7 +11,7 @@ defmodule MozartFetcher.Fetcher do
 
   def process(components) do
     ExMetrics.timeframe "function.timing.fetcher.process" do
-      max_timeout = TimeoutParser.max_timeout(components)
+      max_timeout = TimeoutParser.max(components) + 100
 
       components
       |> Enum.with_index()
--- a/lib/mozart_fetcher/fetcher.ex
+++ b/lib/mozart_fetcher/fetcher.ex
@@ -11,7 +11,9 @@ defmodule MozartFetcher.Fetcher do
 
   def process(components) do
     ExMetrics.timeframe "function.timing.fetcher.process" do
-      max_timeout = TimeoutParser.max(components) + 100
+      max_timeout =
+        TimeoutParser.max(components) +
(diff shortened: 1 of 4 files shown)


lib/mozart_fetcher/fetcher.ex
```diff
@@ -11,7 +11,7 @@ defmodule MozartFetcher.Fetcher do
 
   def process(components) do
     ExMetrics.timeframe "function.timing.fetcher.process" do
-      max_timeout = TimeoutParser.max_timeout(components)
+      max_timeout = TimeoutParser.max(components) + 100
```
ettomatic: I'd probably go for something smaller like `50` but 👍

lib/mozart_fetcher/timeout_parser.ex
```diff
@@ -6,11 +6,13 @@ defmodule MozartFetcher.TimeoutParser do
     query(uri.query)
   end
 
-  def max_timeout(components) do
+  def max(components) do
     components
-    |> Enum.map(&Map.get(&1, :endpoint))
-    |> Enum.map(fn endpoint -> parse(endpoint) end)
-    |> Enum.max()
+    |> Enum.reduce(@default_timeout, &component_timeout_or_current_timeout/2)
```
ettomatic: why this?

ettomatic: so we assume that there's a performance gain with no verification.

samfrench: It should be more performant. Isn't this part more of an opinion on the style.

ettomatic: ok

ettomatic: let's merge it

james-bowers: we assumed it wasn't a performance hit before, with no verification

PR #43, opened 2019-05-24
Add error logging for errors and exceptions

samfrench:
To understand why we get 500s from Fetcher since a change introduced the other day, a rescue has been used to be able to log if we get errors or exceptions.

It seems try/catch/rescue is not a pattern that should be used in this. But, to understand why we have 500s it seems rescue is used for unexpected situations and catch is used where there are expected failures.

If we get an exception raised this will have a message so I can do:
```ex
ex.message
```

If we get an error, then this won't have a message as this is likely to be a struct:
```ex
%FunctionClauseError{
 # properties of struct
}
```

For now I log the item as it is received. If this turns out to be an exception this can be made more specific. If it does not have the message property, then this would in turn raise an error.

I tried to log the stacktrace for the exception (before I made it generic) but I got this, so left that out for now:
```ex
Stump.log(:error, %{message: ex.message, stacktrace: System.stacktrace()})

** (Poison.EncodeError) unable to encode value: {ExUnit.Runner, :"-spawn_test_monitor/4-fun-1-", 4, [file: 'lib/ex_unit/runner.ex', line: 306]}
     code: HTTPClient.get(
     stacktrace:
       (poison) lib/poison/encoder.ex:383: Poison.Encoder.Any.encode/2
       (poison) lib/poison/encoder.ex:259: anonymous fn/3 in Poison.Encoder.List.encode/3
       (poison) lib/poison/encoder.ex:260: Poison.Encoder.List."-encode/3-lists^foldr/2-1-"/3
       (poison) lib/poison/encoder.ex:260: Poison.Encoder.List.encode/3
       (poison) lib/poison/encoder.ex:227: anonymous fn/4 in Poison.Encoder.Map.encode/3
       (poison) lib/poison/encoder.ex:228: Poison.Encoder.Map."-encode/3-lists^foldl/2-0-"/3
       (poison) lib/poison/encoder.ex:228: Poison.Encoder.Map.encode/3
       (poison) lib/poison.ex:41: Poison.encode!/2
       (stump) lib/stump.ex:28: Stump.log/2
       (mozart_fetcher) lib/mozart_fetcher/http_client.ex:30: HTTPClient.get/2
       test/http_client_test.exs:112: (test)
```

### DIFF
--- a/lib/mozart_fetcher/http_client.ex
+++ b/lib/mozart_fetcher/http_client.ex
@@ -4,24 +4,30 @@ defmodule HTTPClient do
   alias MozartFetcher.TimeoutParser
 
   def get(endpoint, client \\ client()) do
-    ExMetrics.timeframe "function.timing.http_client.get" do
-      headers = []
+    try do
+      ExMetrics.timeframe "function.timing.http_client.get" do
+        headers = []
 
-      options = [
-        recv_timeout: TimeoutParser.parse(endpoint),
-        ssl: MozartFetcher.request_ssl(),
-        hackney: [pool: :origin_pool]
-      ]
+        options = [
+          recv_timeout: TimeoutParser.parse(endpoint),
+          ssl: MozartFetcher.request_ssl(),
+          hackney: [pool: :origin_pool]
+        ]
 
-      case make_request(sanitise(endpoint), headers, options, client) do
-        {:error, %HTTPoison.Error{reason: :closed}} ->
-          ExMetrics.increment("http.component.retry")
-          make_request(sanitise(endpoint), headers, options, client)
+        case make_request(sanitise(endpoint), headers, options, client) do
+          {:error, %HTTPoison.Error{reason: :closed}} ->
+            ExMetrics.increment("http.component.retry")
+            make_request(sanitise(endpoint), headers, options, client)
 
-        {k, resp} ->
-          {k, resp}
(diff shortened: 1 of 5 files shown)


woodyblah:
Would it make more sense to add an error handling plug to Cowboy so that we can let exceptions bubble up from anywhere in the code and handle them at the routing layer? https://hexdocs.pm/plug/Plug.ErrorHandler.html

james-bowers:
I thought that, but then it would fail the entire request, and cannot return any other components that were successful. As maybe the one that fails is not a must succeed, then we've failed the request on something that is not required

james-bowers:
Having said that, it might be a good idea to add that error handler anyway, to catch and log any other errors that occur

samfrench:
I would prefer to have this in these cases as we didn't have this before the timeout changes the other day. I am happy to look at having an error plug at a later date, but that may not catch the errors as we think.

Doing reading yesterday, it is impossible to catch errors when the process dies when using async and await. I am not sure if an error plug would be able to catch this.

ettomatic:
I think the ErrorHandler plug is a good idea but for the short term, by rescuing exceptions closer to where we suspect this issue can be,  could make easier to investigate and if needed incrementally add additional logic/logs.

So I'd go with the plug, but probably in a second PR.

PR #45, opened 2019-05-26
Use task stream

ettomatic:
This fixes two issues with the current components fetching approach.

1. starting multiple tasks with `Task.async/1` has all-or-nothing semantics. The crash of a single task takes down all other tasks.

2. `Task.async/1` links the new task to the starter process. Therefore, if any task process crashes, the starter process (`MozartFetcher.Router`) will crash too. This brings to errors like:
```elixir
Server: www-data.com:80 (http)
Request: POST /collect
** (exit) exited in: Task.await(%Task{owner: #PID<0.30047.0>, pid: #PID<0.30094.0>, ref: #Reference<0.3297827767.115605506.186927>}, 3100)
    ** (EXIT) time out
#PID<0.30580.0> running MozartFetcher.Router (connection #PID<0.26472.0>, stream id 36) terminated
Server: www-data.com:80 (http)
Request: POST /collect
** (exit) exited in: Task.await(%Task{owner: #PID<0.30580.0>, pid: #PID<0.30627.0>, ref: #Reference<0.3297827767.115605506.191539>}, 3100)
    ** (EXIT) time out
```

The new proposed approach uses [Task.async_stream](https://hexdocs.pm/elixir/Task.html#async_stream/3). On a first iteration, it doesn't fix the timeout issue, but at least the process is now supervised:

```elixir
defmodule Processor do
  def process(number) do
    :timer.sleep(6000)
  end
end

1..10
|> Task.async_stream(&Processor.process/1)
|> Enum.to_list()

(exit) exited in: Task.Supervised.stream(5000)
    ** (EXIT) time out
    (elixir) lib/task/supervised.ex:276: Task.Supervised.stream_reduce/7
    (elixir) lib/enum.ex:3015: Enum.reverse/1
    (elixir) lib/enum.ex:2649: Enum.to_list/1
```

The timeout is now easily fixable with:
```elixir
defmodule Processor do
  def process(number) do
    :timer.sleep(10000 - (number * 1000))
    number
  end
end

1..10
|> Task.async_stream(&Processor.process/1, on_timeout: :kill_task, max_concurrency: 40)
|> Enum.to_list()

[
  exit: :timeout,
  exit: :timeout,
  exit: :timeout,
  exit: :timeout,
  exit: :timeout,
  ok: 6,
  ok: 7,
  ok: 8,
  ok: 9,
  ok: 10
]
```

### DIFF
--- a/lib/mozart_fetcher/fetcher.ex
+++ b/lib/mozart_fetcher/fetcher.ex
@@ -12,18 +12,18 @@ defmodule MozartFetcher.Fetcher do
     {:error}
   end
 
-  def process(components, buffer \\ @timeout_buffer) do
+  def process(configs, buffer \\ @timeout_buffer) do
     ExMetrics.timeframe "function.timing.fetcher.process" do
-      max_timeout = TimeoutParser.max(components) + buffer
+      max_timeout = TimeoutParser.max(configs) + buffer
 
       stream_opts = [timeout: max_timeout,
                      on_timeout: :kill_task,
                      max_concurrency: @max_concurrency]
 
-      components
+      configs
       |> Enum.with_index()
       |> Task.async_stream(&Component.fetch/1, stream_opts)
-      |> zip(components)
+      |> zip(configs)
       |> decorate_response()
       |> Jason.encode!()
     end
--- a/lib/mozart_fetcher/fetcher.ex
+++ b/lib/mozart_fetcher/fetcher.ex
@@ -47,6 +47,7 @@ defmodule MozartFetcher.Fetcher do
                         reason: reason,
                         id: id})
 
-    %{envelope: %Envelope{}, id: id, index: nil, status: nil}
+    status = if reason == :timeout, do: 408, else: 500
+    %{envelope: %Envelope{}, id: id, index: nil, status: status}
   end
 end
--- a/lib/mozart_fetcher/fetcher.ex
+++ b/lib/mozart_fetcher/fetcher.ex
@@ -12,11 +12,12 @@ defmodule MozartFetcher.Fetcher do
   def process(components) do
     ExMetrics.timeframe "function.timing.fetcher.process" do
       max_timeout = TimeoutParser.max(components)
+      stream_opts = [timeout: max_timeout, on_timeout: :kill_task, max_concurrency: 40]
 
       components
       |> Enum.with_index()
-      |> Task.async_stream(&Component.fetch/1, timeout: max_timeout,  on_timeout: :kill_task, max_concurrency: 40)
-      |> Enum.to_list
+      |> Task.async_stream(&Component.fetch/1, stream_opts)
+      |> Enum.to_list()
       |> extract_successful
       |> decorate_response
       |> Jason.encode!()
@@ -27,14 +28,18 @@ defmodule MozartFetcher.Fetcher do
     %{components: envelopes}
   end
 
+  # returns the successful one (including 500 etc) and logs
+  # the tasks which have not completed.
+  # At the moment it's in form  of `[exit: :timeout]` so hard to
+  # easily log which component has failed.
   defp extract_successful(components) do
     {oks, errors} = Enum.split_with(components, fn {k, _v} -> k == :ok end)
 
     errors |> Enum.each(&log(&1))
-    oks |> Enum.map(fn({:ok, result}) -> result end)
+    oks |> Enum.map(fn {:ok, result} -> result end)
   end
 
   defp log({type, error}) do
-    Stump.log(:error, %{message: "Component Process Error", type: type, error: error })
+    Stump.log(:error, %{message: "Component Process Error", type: type, error: error})
   end
 end
--- a/lib/mozart_fetcher/fetcher.ex
+++ b/lib/mozart_fetcher/fetcher.ex
@@ -23,7 +23,6 @@ defmodule MozartFetcher.Fetcher do
       components
       |> Enum.with_index()
       |> Task.async_stream(&Component.fetch/1, stream_opts)
-      |> validate()
       |> zip(components)
       |> decorate_response()
       |> Jason.encode!()
@@ -34,13 +33,9 @@ defmodule MozartFetcher.Fetcher do
     %{components: envelopes}
   end
 
-  defp validate(responses) do
-    Enum.map(responses, &handle/1)
-  end
-
-  def zip(responses, configs) do
+  defp zip(responses, configs) do
     for {response, config} <- Enum.zip(responses, configs) do
-      Map.merge(response, %{id: config.id})
+      handle(response) |> Map.merge(%{id: config.id})
     end
   end
 
--- a/lib/mozart_fetcher/fetcher.ex
+++ b/lib/mozart_fetcher/fetcher.ex
@@ -3,16 +3,22 @@ defmodule MozartFetcher.Fetcher do
 
   use ExMetrics
 
+  @timeout_buffer 50
+  @max_concurrency 50
+
   def process([]) do
     ExMetrics.increment("error.empty_component_list")
     Stump.log(:error, %{message: "Error cannot process empty component list"})
     {:error}
   end
 
-  def process(components) do
+  def process(components, buffer \\ @timeout_buffer) do
     ExMetrics.timeframe "function.timing.fetcher.process" do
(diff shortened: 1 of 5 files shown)


lib/mozart_fetcher/fetcher.ex
```diff
@@ -11,14 +11,14 @@ defmodule MozartFetcher.Fetcher do
 
   def process(components) do
     ExMetrics.timeframe "function.timing.fetcher.process" do
-      max_timeout =
-        TimeoutParser.max(components) +
-          Application.get_env(:mozart_fetcher, :additional_task_await_timeout)
+      max_timeout = TimeoutParser.max(components)
+      stream_opts = [timeout: max_timeout, on_timeout: :kill_task, max_concurrency: 40]
```
ettomatic: `max_concurrency: 40` is obviously a random number. if not passed it defaults to the number of cores.

ettomatic: Wondering if in the previous approach we were creating a number of processes equivalent to the size of the `component` list. Maybe for huge pages, the number of concurrent requests would cause the timeout issue? We could probably set max_concurrency with different values and load test this...

samfrench: Sometimes we have 50+ components.

samfrench: Should we add a buffer to the timeout?

lib/mozart_fetcher/fetcher.ex
```diff
@@ -11,20 +11,35 @@ defmodule MozartFetcher.Fetcher do
...
       |> Jason.encode!()
     end
   end
 
   defp decorate_response(envelopes) do
     %{components: envelopes}
   end
+
+  # returns the successful one (including 500 etc) and logs
+  # the tasks which have not completed.
```
samfrench: Can we return an empty envelope If they haven't completed, a bit like requester does, or just map the id to a 408 or 500 status as we don't have any content.

I was also thinking about the ordering, that we return an index for the component, but if we are missing some, I assume they will just not be on the page. I was wondering if there were any other issues with that as it is different behaviour from previously.

lib/mozart_fetcher/fetcher.ex
```diff
@@ -11,20 +11,35 @@ defmodule MozartFetcher.Fetcher do
...
+  # At the moment it's in form  of `[exit: :timeout]` so hard to
+  # easily log which component has failed.
+  defp extract_successful(responses) do
+    {oks, errors} = Enum.split_with(responses, fn {k, _v} -> k == :ok end)
+
+    errors |> Enum.each(&log(&1))
+    oks |> Enum.map(fn {:ok, result} -> result end)
+  end
+
+  defp log({type, error}) do
```
samfrench: Can we have a test for this part.
Does `[exit: :timeout]` match the matcher `{type, error}`? One is a tuple and the other a list. It might be my knowledge about this isn't great, so it might do!

This should be `{exit: reason}` according to the docs and the comment further up. So type would be "exit" and reason would be error here? Maybe it's just naming that makes it a bit confusing.

lib/mozart_fetcher/fetcher.ex
```diff
@@ -11,20 +11,35 @@ defmodule MozartFetcher.Fetcher do
...
   end
+
+  # returns the successful one (including 500 etc) and logs
+  # the tasks which have not completed.
+  # At the moment it's in form  of `[exit: :timeout]` so hard to
+  # easily log which component has failed.
+  defp extract_successful(responses) do
+    {oks, errors} = Enum.split_with(responses, fn {k, _v} -> k == :ok end)
+
+    errors |> Enum.each(&log(&1))
```
samfrench: Do we need to do an each here if it's only errors as we could pass all and log all at once?

lib/mozart_fetcher/fetcher.ex
ettomatic: the size of the batches is hard to define, but could be the root of the previous issue, if we process a page made of say 80 components they will be processed concurrently but given a limited number of cores not in parallel, this could possibly add additional latency to the request? The batches can alleviate this problem as the timeout will be measured against the current batch.
 
ok for the additional buffer! I'd probably add a few milliseconds

lib/mozart_fetcher/fetcher.ex
```diff
@@ -1,30 +1,46 @@
...
+
+  defp validate(responses) do
+    Enum.map(responses, &handle/1)
+  end
+
+  defp handle({:ok, result}), do: result
+
+  defp handle({state, reason}) do
+    Stump.log(:error, %{message: "Component Process Error", state: state, reason: reason})
+    %Envelope{}
```
samfrench: Should we fill this with a 408 or 500 status code?

samfrench: In Requester, it adds an error key to the envelope...
```go
Envelope: aggregator.Envelope{Error: err.Error()},
```

ettomatic: ah, is it needed? Are you suggesting to add a fake `500` status code to the empty struct?

samfrench: Possibly. Or a 408 if we didn't get a response. I don't know if we should have error as that wasn't part of the envelope originally. We check if status is 200 so it doesn't really matter if we have it or not. The same with the error, that we don't do anything with it outside of this application.

samfrench:
I think we should merge this and get this onto test tomorrow and run a load test(s) against this. If all is good, let's put this live either first thing Thursday or first thing Monday.

lib/mozart_fetcher/fetcher.ex
ettomatic: I think I'd prefer for now to avoid any mocked status. My only concern is that the empty envelope will not have a status key at all, this could possibly break Assembler?

samfrench: I see. I thought we had the same keys, such as id, status, index and the envelope and status were empty. I looked at this bit earlier and we check for a 200, so as long as it's safe, this will be fine.

The part that now seems an issue is the id. It looks like we raise an error if we don't get a matching id:
> https://github.com/bbc/mozart-assembler/blob/master/src/lib/mozart_assembler/validators/components.rb#L32

samfrench: We should change the logic that checks this, _nil if valid and not required, exception if not valid, or the component if successful..._

lib/mozart_fetcher/fetcher.ex
```diff
@@ -1,30 +1,52 @@
...
+
+  defp handle({:ok, result}, _id), do: result
+
+  defp handle({state, reason}, id) do
+    Stump.log(:error, %{message: "Component Process Error",
+                        state: state,
+                        reason: reason,
+                        id: id})
+
+    %{envelope: %Envelope{}, id: id, index: nil, status: nil}
```
samfrench: This would allow us to put a status, but we can think about this another time if needed.

test/fetcher_test.exs
```diff
@@ -11,6 +11,30 @@ defmodule MozartFetcher.FetcherTest do
...
+  end
+
+  test "it returns empty envelopes for components erroring out with no generated response" do
+    config = [
+      %Config{endpoint: "http://localhost:8082/foo", id: "foo"},
+      %Config{endpoint: "http://localhost:8082/timeout", id: "timeout"}
+    ]
+
+    expected =
+      ~s({"components":[{"envelope":{"bodyInline":"","bodyLast":[],"head":[]},"id":"foo","index":0,"status":200},{"envelope":{"bodyInline":"","bodyLast":[],"head":[]},"id":"timeout","index":null,"status":null}]})
```
samfrench: Why do we lose the index for the empty ones?

lib/mozart_fetcher/fetcher.ex
ettomatic: I could easily add a bit of logic which will inject a 408 id reason is `:timeout` otherwise 500.

ettomatic: I'm doing that now

test/fetcher_test.exs
ettomatic: the order is guaranteed but we don't have the `index` value in the configs. I don't think it's necessary. Otherwise, it requires an additional Enumeration.

PR #46, opened 2019-06-05
Add gzip header on request and uncompress response

samfrench:
It seems other libraries (such as Go) automatically add an "accept-encoding" header when making requests and these uncompress the responses. This header means a server can respond with compressed content based on the value and will add a "content-encoding" header on the response for the client.

HTTPoison does not do this. There is an issue in Hackney open from 2015 to do this. For now, I have added a "gzip" header when making the requests and uncompress the response.

### DIFF
--- a/lib/mozart_fetcher/http_client.ex
+++ b/lib/mozart_fetcher/http_client.ex
@@ -6,7 +6,7 @@ defmodule HTTPClient do
   def get(endpoint, client \\ client()) do
     try do
       ExMetrics.timeframe "function.timing.http_client.get" do
-        headers = []
+        headers = [{"accept-encoding", "gzip"}]
 
         options = [
           recv_timeout: TimeoutParser.parse(endpoint),
@@ -20,7 +20,7 @@ defmodule HTTPClient do
             make_request(sanitise(endpoint), headers, options, client)
 
           {k, resp} ->
-            {k, resp}
+            handle_response({k, resp})
         end
         |> log_errors_and_return()
       end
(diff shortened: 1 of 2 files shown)


lib/mozart_fetcher/http_client.ex
```diff
@@ -39,6 +39,40 @@ defmodule HTTPClient do
     client.get(endpoint, headers, options)
   end
 
+  defp handle_response({:ok, response}) do
+    {:ok,
+     %HTTPoison.Response{
```
james-bowers: how come we don't use something like [`Map.put`](https://hexdocs.pm/elixir/Map.html#put/3) to update the body instead of rebuilding the response struct?

lib/mozart_fetcher/http_client.ex
```diff
@@ -39,6 +39,40 @@ defmodule HTTPClient do
...
+       body:
+         decode_response_body(response.body, content_encoding(process_headers(response.headers))),
+       status_code: response.status_code
+     }}
+  end
+
+  defp handle_response(response), do: response
+
+  defp process_headers(headers) do
+    Enum.map(headers, fn {k, v} -> {String.downcase(k), v} end)
```
james-bowers: If we are using http1, does the casing of the response headers matter?

lib/mozart_fetcher/http_client.ex
woodyblah: Yep, rebuilding a HTTPoison.Response binds us to HTTPoison, making migrating to a different web client library harder

lib/mozart_fetcher/http_client.ex
```diff
@@ -39,6 +39,40 @@ defmodule HTTPClient do
...
+  end
+
+  defp decode_response_body(body, "gzip"), do: :zlib.gunzip(body)
+  defp decode_response_body(body, "x-gzip"), do: :zlib.gunzip(body)
```
james-bowers: not a big deal at all, I just expected these functions to do different things. Maybe it'd read better having a guard to check the content type?

lib/mozart_fetcher/http_client.ex
woodyblah: These are response headers, so I think it's just normalising them for this line: https://github.com/bbc/mozart-fetcher/pull/46/files#diff-d515d4ebcd8ec474ea1d0cfcdc0c96c7R73

lib/mozart_fetcher/http_client.ex
```diff
@@ -39,6 +39,40 @@ defmodule HTTPClient do
...
+      {_, content_encoding} ->
+        content_encoding
+
+      nil ->
+        ""
+    end
+  end
+
+  defp get_content_encoding(headers) do
+    List.keyfind(headers, "content-encoding", 0)
```
james-bowers: nice

woodyblah:
I would really like tests for something like this. As James mentioned you could use the fake origin https://github.com/bbc/mozart-fetcher/blob/master/lib/mozart_fetcher/fake_origin.ex
Paths could be added that returns gzipped content and the content-encoding header.

Or you could follow the approach of some of the other tests that are sending a mock client in the params, and create one that returns a response struct with gzipped content in the body.

lib/mozart_fetcher/http_client.ex
samfrench: I did think there would be a better way. Thank you. I have added `Map.put` now.

lib/mozart_fetcher/http_client.ex
samfrench: It matters when I am looking for a specific header.

I see HTTPoison has callbacks for response headers and body, but there didn't seem to be one for both, so I left that out for now.

lib/mozart_fetcher/http_client.ex
samfrench: I could do. I'll leave this for now as this will be refactored into separate modules in the near future.

samfrench:
@woodyblah, I was thinking earlier about using the mock client like other tests do. I have added one using that approach as I prefer that.

samfrench:
@woodyblah, the refactoring planned is to separate HTTPoison and the client, so we can swap out HTTPoison easily. Currently we are quite tied to this in matching the response types it provides. At least it's all kept in the client for now.

PR #48, opened 2019-06-06
Log simple string and metric for when component fetching errors

samfrench:
The exception wasn't able to be converted to JSON and produced an error when logging.

### DIFF
--- a/lib/mozart_fetcher/http_client.ex
+++ b/lib/mozart_fetcher/http_client.ex
@@ -25,8 +25,9 @@ defmodule HTTPClient do
         |> log_errors_and_return()
       end
     rescue
-      ex ->
-        Stump.log(:error, %{message: ex})
+      _ ->
+        ExMetrics.increment("http.component.error")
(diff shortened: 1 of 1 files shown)


lib/mozart_fetcher/http_client.ex
```diff
@@ -25,8 +25,9 @@ defmodule HTTPClient do
         |> log_errors_and_return()
       end
     rescue
-      ex ->
-        Stump.log(:error, %{message: ex})
+      _ ->
+        ExMetrics.increment("http.component.error")
```
ettomatic: could you please add the error you got in the logs

ettomatic: just to explain why you did this

samfrench: I have added it to the description.

woodyblah:
I think it would make sense if Stump was able to handle it's own Poison encoding errors (out of the scope of this fix of course, I'll write a ticket for it)

PR #54, opened 2019-12-17
RESFRAME 3032: invalid envelope format

JoeARO:
This PR ensures that we do not raise a KeyError when trying to convert a map into a struct instead we check for the key before the conversion and handle accordingly.

Also contains some formatting of the rest of the project

### DIFF
--- a/lib/mozart_fetcher/decoder.ex
+++ b/lib/mozart_fetcher/decoder.ex
@@ -1,8 +1,14 @@
 defmodule MozartFetcher.Decoder do
   def data_to_struct(data, struct) do
     case Jason.decode(data, keys: :atoms) do
-      {:ok, decoded} -> {:ok, to_struct(decoded, struct)}
-      {:error, _}    -> {:error}
+      {:ok, decoded} ->
+        case to_struct(decoded, struct) do
+          {:error, _} -> {:error}
+          struct -> {:ok, struct}
+        end
+
+      {:error, _} ->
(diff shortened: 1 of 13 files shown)


lib/mozart_fetcher/decoder.ex
```diff
@@ -1,27 +1,42 @@
 defmodule MozartFetcher.Decoder do
   def data_to_struct(data, struct) do
-    case Jason.decode(data, keys: :atoms) do
-      {:ok, decoded} -> {:ok, to_struct(decoded, struct)}
-      {:error, _}    -> {:error}
+    # case Jason.decode(data, keys: :atoms) do
```
woodyblah:
```suggestion
```

I think this comment can go

lib/mozart_fetcher/decoder.ex
```diff
@@ -1,27 +1,41 @@
...
 
-  defp to_struct(map, struct) do
-    case struct!(struct, map) do
-      {:error} ->
+  def to_struct(map, struct) do
+    case Map.keys(Map.from_struct(struct)) |> Enum.all?(&Map.has_key?(map, &1)) do
+      true ->
+        case struct(struct, map) do
+          struct -> {:ok, struct}
+        end
```
woodyblah: This probably doesn't need to be a `case` statement.

`true -> {:ok, struct(struct, map)}`

should be enough

PR #57, opened 2020-01-06
run tests in Jenkins job

james-bowers:


Jenkinsfile
```diff
@@ -39,6 +39,14 @@ node {
       git url: 'https://github.com/bbc/mozart-fetcher-build', credentialsId: 'github', branch: 'master'
     }
   }
+
+  stage('Run tests') {
+    docker.image('qixxit/elixir-centos').inside("-u root -e MIX_ENV=test") {
+      sh 'mix deps.get'
```
samfrench: Should we do a stash like we do in Mozart? There is a slight chance (very unlikely) that the dependencies we test with could be different from the ones we package in the release.

woodyblah: `mix deps.get` does get run again with `MIX_ENV=prod` here: https://github.com/bbc/mozart-fetcher/pull/57/files#diff-58231b16fdee45a03a4ee3cf94a9f2c3L45-L46

Perhaps the build directory could be deleted just to make sure, but a stash isn't necessary as that is only used by the ruby Jenkins job to make sure the deps are available even if the stages run on different nodes (which doesn't happen for scripted pipelines like this one)

james-bowers: I don't think we have to tbh. The behaviour of the app, (runtime and compile time) is different between mix environments (test, dev and prod) due to things like testing with `Mox` and other config values in the `config/<<mix env>>.exs` files.

In my opinion, I think we are more likely to have an error in our own config between mix envs that the `mix test` won't catch, than have an error caused by potentially installing different minor versions between testing and building the production release.

samfrench: Interestingly it seems we'd want to install dependencies as prod so they're there for when we build and package the app. But we'd want to run the tests as test... Alternatively we install them twice, but there could be differences between test and prod. Happy to consider this later as we weren't running any of the tests before this.

PR #58, opened 2020-01-15
remove validation for config and stricter error catching for decode

JoeARO:
Tested this on local mozart set up and it fixes the decoding error introduced in this https://github.com/bbc/mozart-fetcher/pull/54

Screenshot 2020-01-16 at 13 03 41
Screenshot 2020-01-16 at 13 04 02

### DIFF
--- a/lib/mozart_fetcher/decoder.ex
+++ b/lib/mozart_fetcher/decoder.ex
@@ -4,7 +4,7 @@ defmodule MozartFetcher.Decoder do
   def decode_envelope(data, struct) do
     case Jason.decode(data, keys: :atoms) do
       {:ok, decoded} ->
-        to_struct(decoded, struct)
+        to_envelope(decoded, struct)
 
       {:error, error = %Jason.DecodeError{}} ->
         Stump.log(:error, "Envelope decode error: #{error.data}")
@@ -29,7 +29,7 @@ defmodule MozartFetcher.Decoder do
     end)
   end
 
-  defp to_struct(map, struct = %Envelope{}) do
+  defp to_envelope(map, struct = %Envelope{}) do
     try do
       {:ok, struct!(struct, map)}
     rescue
--- a/lib/mozart_fetcher/decoder.ex
+++ b/lib/mozart_fetcher/decoder.ex
@@ -2,62 +2,46 @@ defmodule MozartFetcher.Decoder do
   alias MozartFetcher.{Envelope, Config}
 
   def decode_envelope(data, struct) do
-    with {:ok, decoded} <- Jason.decode(data, keys: :atoms),
-         {:ok, struct} <- to_struct(decoded, struct) do
-      {:ok, struct}
-    else
+    case Jason.decode(data, keys: :atoms) do
+      {:ok, decoded} ->
+        to_struct(decoded, struct)
+
       {:error, error = %Jason.DecodeError{}} ->
         Stump.log(:error, "Envelope decode error: #{error.data}")
-        {:error}
-
-      {:error, "Failed to convert map to struct"} ->
         {:error}
     end
   end
 
-  # This could be more generic if we add a list_name then Enum.map(data[list_name]) instead of specifically for components here
   def decode_config(data, struct) do
-    with {:ok, decoded} <- Jason.decode(data, keys: :atoms),
-         {:ok, decoded} <- components_to_struct(decoded[:components], struct) do
-      {:ok, decoded}
-    else
+    case Jason.decode(data, keys: :atoms) do
+      {:ok, decoded} ->
+        {:ok, components_to_struct(decoded[:components], struct)}
+
       {:error, error = %Jason.DecodeError{}} ->
         Stump.log(:error, "Config decode error: #{error.data}")
-        {:error}
-
-      {:error, "Invalid Config list"} ->
         {:error}
     end
   end
 
   defp components_to_struct(components, struct) do
-    component_list =
-      Enum.map(components, fn map ->
-        struct(struct, map)
-      end)
-
-    case Enum.any?(component_list, fn component ->
-           component != {:error}
-         end) do
-      true -> {:ok, component_list}
-      false -> {:error, "Invalid Config list"}
-    end
+    Enum.map(components, fn map ->
+      struct(struct, map)
+    end)
   end
 
   defp to_struct(map, struct = %Envelope{}) do
-    case struct(struct, map) do
-      {:error} ->
-        ExMetrics.increment("error.components.decode")
+    try do
+      {:ok, struct!(struct, map)}
+    rescue
+      KeyError ->
+        ExMetrics.increment("error.envelope.decode")
 
         Stump.log(:error, %{
           message: "Map contains invalid keys cannot convert to Envelope",
           map: map
         })
 
-        {:error, "Failed to convert map to struct"}
-
-      struct ->
-        {:ok, struct}
+        {:ok, %Envelope{}}
     end
   end
 end
--- a/lib/mozart_fetcher/decoder.ex
+++ b/lib/mozart_fetcher/decoder.ex
@@ -10,7 +10,7 @@ defmodule MozartFetcher.Decoder do
         Stump.log(:error, "Envelope decode error: #{error.data}")
         {:error}
 
-      _ ->
+      {:error, "Failed to convert map to struct"} ->
         {:error}
     end
   end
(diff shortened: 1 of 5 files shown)


test/decoder_test.exs
```diff
@@ -64,38 +65,6 @@ defmodule MozartFetcher.DecoderTest do
 
       assert Decoder.decode_config(json, %Config{}) == expected
     end
-
-    test "when a map in the list of JSON has an invalid key, the other components are still returned" do
```
james-bowers: instead of removing this test, do we want to test the new behaviour when a component is invalid?

lib/mozart_fetcher/decoder.ex
```diff
@@ -6,58 +6,81 @@ defmodule MozartFetcher.Decoder do
          {:ok, struct} <- to_struct(decoded, struct) do
       {:ok, struct}
     else
-      _ -> {:error}
+      {:error, error = %Jason.DecodeError{}} ->
+        Stump.log(:error, "Envelope decode error: #{error.data}")
+        {:error}
+
+      _ ->
```
james-bowers: do you know what other values we could get here?

lib/mozart_fetcher/decoder.ex
```diff
@@ -6,58 +6,81 @@ defmodule MozartFetcher.Decoder do
...
+      {:ok, decoded}
+    else
+      {:error, error = %Jason.DecodeError{}} ->
+        Stump.log(:error, "Config decode error: #{error.data}")
+        {:error}
+
+      {:error, "Invalid Config list"} ->
+        {:error}
+
+      _ ->
```
james-bowers: do you know what other values we could get here? :)

test/decoder_test.exs
JoeARO: Ahh yes I forgot about that, was too eager to check it worked aha

lib/mozart_fetcher/decoder.ex
```diff
@@ -2,62 +2,46 @@ defmodule MozartFetcher.Decoder do
...
   end
 
-  def to_struct(map, struct = %Envelope{}) do
-    case [:head, :bodyInline, :bodyLast] |> Enum.all?(&Map.has_key?(map, &1)) do
-      true ->
-        {:ok, struct(struct, map)}
-
-      false ->
-        ExMetrics.increment("error.components.decode")
+  defp to_struct(map, struct = %Envelope{}) do
```
james-bowers: `to_struct` sounds too generic to me, how about `to_envelope`?

JoeARO: ahh yes good suggestion, I like it, I like it a lot!

lib/mozart_fetcher/decoder.ex
```diff
@@ -2,62 +2,46 @@ defmodule MozartFetcher.Decoder do
...
+    try do
+      {:ok, struct!(struct, map)}
+    rescue
```
james-bowers: as function is only used for the `Envelope` struct, I think this try and rescue could be replaced by checking the keys instead? I think that'd be more Elixir like.

Also,[ I believe this `case` is now redundant](https://github.com/bbc/mozart-fetcher/blob/master/lib/mozart_fetcher/envelope.ex#L8) because we will always return `{:ok, %Envelope{}}`. And therefore i[ncorrectly increment this metric ](https://github.com/bbc/mozart-fetcher/blob/master/lib/mozart_fetcher/envelope.ex#L10)

lib/mozart_fetcher/router.ex
```diff
@@ -47,7 +47,7 @@ defmodule MozartFetcher.Router do
       {:error} ->
         ExMetrics.increment("error.components.decode")
         Stump.log(:error, %{message: "Failed to decode components into list"})
-        {:error}
+        :error
```
boonious: What is the reason behind this change? This is inconsistent with the `{:ok, _}` and {:error, _} cases in this and other modules. Could return a {error, exception} that provide the error messages for the preceding `send_response/2`, rather than for it to be hard-coded as below:

```
 def send_response(:error, conn) do
    conn
    |> put_resp_content_type("application/json")
    |> send_resp(500, "Internal Server Error")
  end
```
To 

```
 def send_response({:error, exception}, conn) do
    conn
    |> put_resp_content_type("application/json")
    |> send_resp(exception.status, exception.message)
  end
```

Could reuse existing exceptions (500 from HTTP clients) for other library or define one with `defexception`.

test/decoder_test.exs
```diff
@@ -20,15 +20,16 @@ defmodule MozartFetcher.DecoderTest do
...
 
-    test "when the JSON contains invalid keys we rescue and return an error" do
-      json = ~s({"head":[],"bodyInLine":"<DIV id=\\"site-container\\">","bodyLast":[]})
+    test "when the %Envelope{} has invalid keys we still return an empty %Envelope" do
+      json = ~s({"head":[],"foobar":"<DIV id=\\"site-container\\">","bodyLast":[]})
 
-      assert Decoder.decode_envelope(json, %Envelope{}) == {:error}
+      assert Decoder.decode_envelope(json, %Envelope{}) ==
+               {:ok, %MozartFetcher.Envelope{bodyInline: "", bodyLast: [], head: []}}
     end
```
boonious: In the statement:  `%Envelope{}` (struct) has invalid keys, but in test it's referring to the incoming JSON body?

test/decoder_test.exs
```diff
@@ -20,15 +20,16 @@ defmodule MozartFetcher.DecoderTest do
...
   end
 
-  describe "#list_to_struct_list" do
-    test "when the list of JSON is successfuly decoded it is transformed into a list of structs" do
+  describe "#decode_config" do
+    test "when the %Config{} is valid JSON it returns a list of Config structs" do
```
boonious: `%Config{}` **isn't** a valid JSON : )

test/decoder_test.exs
JoeARO: in the json variable the key `bodyLast` is `foobar` ill try and make it a bit clearer though if you think that is necessary, maybe something like:
`when the json does not conform to the %Envelope{} structs keys we return an error`?

the return value of this is going to change back to {:error} instead of an empty %Envelope{} as the envelope class will return the empty %Envelope{}

JoeARO: `test "when the JSON contains invalid keys for an %Envelope{} we return an error" do`
Gonna go with this thinks its clearer

boonious: Note the test and code logic does not return an error. It returns an empty struct.

boonious: It makes sense to use "the JSON" since you already used that in the preceding case.

test/decoder_test.exs
JoeARO: Have reworded this thanks :)

test/decoder_test.exs
JoeARO: Sorry I think I was a bit confusing in my message, the behaviour has been updated in my latest commits and the wording has been made clearer now. Thanks :)

lib/mozart_fetcher/router.ex
JoeARO: I have followed the approach you have suggested, we dont really update this application much though so I dont think the exception status/message is going to be used massively but I think the main thing to gain from it is setting a good example for others who may look through our code base, thanks for pointing it out!

lib/mozart_fetcher/decoder.ex
JoeARO: I have updated the codebase so that we only log specifically when we are close to the specific reasoning in the decoder and have left it so only the Envelope module will do the metricing otherwise we would end up metricing multiple times as you have suggested.

I think im going to leave the key error as I think its a bit clearer, I think it has a bit too much logic with the key checking unless someone has a better suggestion than this to check them:
`case [:head, :bodyInline, :bodyLast] |> Enum.all?(&Map.has_key?(map, &1)) do`

lib/mozart_fetcher/router.ex
boonious: Thanks Joe, that's a pragmatic solution. A good starting point for more elaborated error handling in the future if needs be.

lib/mozart_fetcher/decoder.ex
JoeARO: Also have changed the decoder to return `{:error}` to keep things consistent and so that it is the Envelope module that returns the empty %Envelope{} on error.

PR #61, opened 2020-11-04
RESFRAME-2740 - Remove Envelope validation

woodyblah:
RESFRAME-2740

This PR removes the Envelope struct and any validation/decoding/builders based around that. This PR is the first of three stages to bring the handling of abstract JSON data structures (from ares specifically) and Envelope format json data structures.

This PR allows any data structure to be returned to Assembler under the `envelope:` key of each component. This will enable a Second PR in Assembler to remove the `format: 'ares'` flag that it sends to fetcher, which will then allow a third PR to finally remove all specific ares logic from the fetcher, and have both envelope and ares data be handled in the same way.


lib/mozart_fetcher/component.ex
```diff
@@ -26,8 +26,15 @@ defmodule MozartFetcher.Component do
...
          {:ok, %HTTPoison.Response{status_code: 200, body: body}}
        ) do
-    metric(config.id, endpoint, 200)
-    %{index: component_index, id: config.id, status: 200, envelope: Envelope.build(body)}
+    case Jason.decode(body, keys: :atoms) do
+      {:ok, data} ->
+        metric(config.id, endpoint, 200)
+        %{index: component_index, id: config.id, status: 200, envelope: data}
+
+      {:error, _err} ->
```
samfrench, line 34: Previously we logged when this errored. Is this something we need or found useful before?

lib/mozart_fetcher/component.ex
```diff
@@ -26,8 +26,15 @@ defmodule MozartFetcher.Component do
...
+        metric(config.id, endpoint, :json_decode_error)
+        %{index: component_index, id: config.id, status: 200, envelope: %{}}
```
samfrench, line 36: Previously we returned an empty envelope so the necessary keys were present. If it's an empty map would that cause issues if the keys do not exist? I know assembler validates against the keys, but wondering if an empty envelope would have hidden optional components not being fetched whereas we would now error due to the keys not existing.

samfrench:
Do we fetch multiple non-envelope components in one request? Wondering if there would be less changes if we fetched non-envelope things from Assembler or directly.

This could be a good tidy up of the logic and makes things simpler, but most things usually exist for a reason. Previously we defaulted with empty envelopes and now we don't have that structure. I haven't looked into it at all, so could all be fine and not something we want to concern ourselves with.

lib/mozart_fetcher/component.ex
woodyblah: The `metric` method called below is perhaps slightly misleading as it will also log for errors. Although the message will be slightly different it does mean it will still log the errors

lib/mozart_fetcher/component.ex
woodyblah: The keys not existing doesn't cause an error anywhere in assembler, although it does cause the component to be marked as not valid as this line will return false: https://github.com/bbc/mozart-assembler/blob/master/src/lib/mozart_assembler/validators/envelope.rb#L17

If anything I think this is safer. Currently if a component endpoint was responding with a 200 and some invalid JSON then fetcher would return an empty envelope with the correct keys and a 200 would be considered "valid" in assembler and therefore the empty content merged into the template. For non-must-succeed components this makes no difference, if an empty content is merged in, or an invalid component is left out equals the same behaviour. In the case of must-succeed components though this new behaviour is preferable I think, as before the page would be rendered but with blank content taking the place of the must-succeed components, whereas now it will correctly 500.

woodyblah:
> Do we fetch multiple non-envelope components in one request?

I'm not sure actually, but if not then it could be easier just to fetch them directly in assembler and remove all non-envelope logic from here. I guess I could query the page_configs and see if any use the `format: ares` flag, if they don't it'll just be the ares component which only fetches a single item iirc

PR #62, opened 2020-11-20
RESFRAME-2740 - Remove all ares specific handling from fetcher

woodyblah:
After https://github.com/bbc/mozart-assembler/pull/355 was merged the Assembler goes directly to Ares for JSON data, so the fetcher no longer needs specific Ares handling logic. 

#61 has been closed in favour of this PR, as there is now no need to bring the envelope and ares handling inline as the fetcher no longer needs to deal with Ares requests.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -36,7 +36,7 @@ node {
   stage('Checkout mozart-fetcher-build') {
     sh 'mkdir -p mozart-fetcher-build'
     dir('mozart-fetcher-build') {
-      git url: 'https://github.com/bbc/mozart-fetcher-build', credentialsId: 'github', branch: 'master'
+      git url: 'https://github.com/bbc/mozart-fetcher-build', credentialsId: 'de1d9453-493a-4f18-a2ab-507822b96188', branch: 'master'
     }
   }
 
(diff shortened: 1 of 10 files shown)


JoeARO (approved):

Jenkinsfile
```diff
@@ -36,7 +36,7 @@ node {
   stage('Checkout mozart-fetcher-build') {
     sh 'mkdir -p mozart-fetcher-build'
     dir('mozart-fetcher-build') {
-      git url: 'https://github.com/bbc/mozart-fetcher-build', credentialsId: 'github', branch: 'master'
+      git url: 'https://github.com/bbc/mozart-fetcher-build', credentialsId: 'de1d9453-493a-4f18-a2ab-507822b96188', branch: 'master'
```
james-bowers, line 39: Is this to help mitigate the GitHub API rate limiting by using different credentials?

james-bowers (approved):
It's great when we can remove stuff

Jenkinsfile
woodyblah: No it's just to get the build to pass, there are no credentials with the id `github` so it was causing failures. Not sure what's changed exactly as it used to work...

PR #63, opened 2021-03-29
Resframe 4190 user agent

JoeARO:


lib/mozart_fetcher/http_client.ex
```diff
@@ -37,7 +37,11 @@ defmodule HTTPClient do
   end
 
   defp make_request(endpoint, headers, options, client) do
-    client.get(endpoint, headers, options)
+    client.get(endpoint, request_headers(headers), options)
+  end
+
+  defp request_headers(headers) do
+    headers ++ [{ 'User-Agent', 'Mozart-Fetcher' }]
```
samfrench, line 44: Can this be "MozartFetcher" as in the ticket.

JoeARO: Apologies will update that now :)

JoeARO: Its cause the ticket title says 'Mozart-Fetcher' but you are right

test/http_client_test.exs
```diff
@@ -34,6 +34,11 @@ defmodule HTTPClientTest do
       assert resp.request_url == "http://localhost:8082/foo%20bar?baz=%20true"
     end
 
+    test "the request headers contain a User-Agent header with the value Mozart-Fetcher" do
```
samfrench, line 37: Do we need to update the description?

samfrench (approved):

test/http_client_test.exs
JoeARO: Done thanks

Raymonoir (approved):

test/http_client_test.exs
```diff
@@ -34,6 +34,11 @@ defmodule HTTPClientTest do
       assert resp.request_url == "http://localhost:8082/foo%20bar?baz=%20true"
     end
 
+    test "the request headers contain a User-Agent header with the value MozartFetcher" do
+      {:ok, resp} = HTTPClient.get("http://localhost:8082/foo/bar")
+      assert Enum.member?(resp.request.headers, {'User-Agent', 'MozartFetcher'}) == true
```
boonious, line 39: The following should also be possible, just to beef up your future options : )

```ex
  assert  {'User-Agent', 'MozartFetcher'} in resp.request.headers
```

boonious (approved):

test/http_client_test.exs
boonious: Also, the following is idiomatic (not needing`== true`) as `Enum.member?/2` is a boolean function

```ex
  assert Enum.member?(resp.request.headers, {'User-Agent', 'MozartFetcher'})
```

Raymonoir (approved):

ettomatic (approved):

PR #64, opened 2021-04-28
RESFRAME-4167 - updated elixir/erlang, used new docker image in jenkins

Raymonoir:
JIRA: https://jira.dev.bbc.co.uk/browse/RESFRAME-4167


ettomatic (approved):

chrishop (approved):

lib/mozart_fetcher/component.ex
```diff
@@ -65,8 +65,6 @@ defmodule MozartFetcher.Component do
   defp metric(id, endpoint, reason) do
     ExMetrics.increment("error.component.process")
     ExMetrics.increment("error.component.process.#{id}")
-    ExMetrics.increment("error.component.process.#{reason}")
```
ettomatic, line 68: I think it makes sense not to metric reason, also wondering about id as I don't think we are really using these metrics anywhere...

Did you say that reason is not available anyomore? Or otherwise what is it, a binary string like `"timeout"`?

Raymonoir: This breaks because the `reason` key within HTTPoison has the type `any()` shown here: https://hexdocs.pm/httpoison/HTTPoison.Error.html, this means reason can be some types which break when places within a string like in the metric. 

I am happy to remove ID as like you say It does not seemed to be used.

ettomatic: but `reason` will still work in the log entry below at line 73?

PR #68, opened 2023-01-13
Update and use to V2 of cfn-signal

JoeARO:
This PR updates the CFN Signal package to v2 as it was not working with metadata v2 options we were introducing with this PR: https://github.com/bbc/frameworks-infra/pull/181


samfrench (approved):

ettomatic (commented):
Shouldn't we also remove the repo from `infrastructure/stacks/repoconfig.json` ?

JoeARO:
> Shouldn't we also remove the repo from `infrastructure/stacks/repoconfig.json` ?

Doesn't exist in this repo

ettomatic (approved):

PR #70, opened 2023-01-18
Use amazon cloudwatch agent

JoeARO:
This PR removes the bbc-statsd cloudwatch agent, configures the amazon cloudwatch agent to collect statsd metrics and adds a global BBCEnvironment tag to metrics emitted from ExMetrics.


rel/runtime_config.exs
```diff
@@ -0,0 +1,15 @@
+use Mix.Config
+
+[
+  "PRODUCTION_ENVIRONMENT"
```
ettomatic, line 4: I guess IDCTA will come at a later PR isn't it?
https://github.com/bbc/belfrage/blob/master/rel/runtime_config.exs#L25

ettomatic (approved):

rel/runtime_config.exs
samfrench: Fetcher doesn't use IDCTA.

rel/runtime_config.exs
```diff
@@ -0,0 +1,15 @@
...
+]
+|> Enum.each(fn config_key ->
+  if System.get_env(config_key) == nil do
+    raise "Config not set in environment: #{config_key}"
+  end
+  config :belfrage_ccp,
+         Keyword.new([{String.to_atom(String.downcase(config_key)), System.get_env(config_key)}])
+end)
```
samfrench, line 12: Same question about this as asked on the other PR.

JoeARO: Its setting the `PRODUCTION_ENVIRONMENT` variable in the list, I just copied what we do there from belfrage/ccp

thymusvulgaris (approved):

samfrench (approved):

PR #73, opened 2023-04-06
use updated Docker image

ettomatic:
https://hub.docker.com/r/ettomatic/elixir-centos


brunoripa (approved):
Not much to see :), in terms of correctness it's ok.

PR #77, opened 2023-09-28
RESFRAME 5444 codebuild warnings

GuiHeurich:
https://jira.dev.bbc.co.uk/browse/RESFRAME-5444

This change solves a warning from Codebuild:

```
warning: File listed twice: /home/component/component-status-cfn-signal.sh
```

I have [deployed this to TEST](https://cosmos.tools.bbc.co.uk/deployments/7967018) and confirmed that `component-status-cfn-signal.sh` can be found in `/home/component`. The component itself `mozart-fetcher` and other files can also be found in the same folder. 

Screenshot 2023-09-29 at 11 48 46

### DIFF
--- a/mozart-fetcher.spec
+++ b/mozart-fetcher.spec
@@ -50,7 +50,6 @@ systemctl enable cloudformation-signal
 /bin/chown -R component:component /var/log/component
 
 %files
-/home/component
 %attr(0755, component, component) /home/component/component-status-cfn-signal.sh
 /usr/lib/systemd/system/cloudformation-signal.service
 /usr/lib/systemd/system/mozart-fetcher.service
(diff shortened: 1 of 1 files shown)


samfrench (approved):

mozart-fetcher.spec
```diff
@@ -50,8 +50,7 @@ systemctl enable cloudformation-signal
 /bin/chown -R component:component /var/log/component
 
 %files
-/home/component
-%attr(0755, component, component) /home/component/component-status-cfn-signal.sh
+%attr(0755, component, component) /home/component/*
```
ettomatic, line 53: so you are resolving the warning by changing the permissions octal value?

GuiHeurich: Not exactly, I don't think. I think what solves the warning is removing `/home/component` from line 53, which lists all files under that folder, including `component-status-cfn-signal.sh`. So, because that file is the only file in the folder, the octal value is the same as it was before.

tracy-o (approved):

ettomatic (approved):

PR #80, opened 2024-04-10
Conditionally add FABL API ctx-unwrapped header to requests to the FABL API

craigkj:
By default the FABL API returns a fixed data response which is not compatible with mozart page composition (it's not envelope format), meaning that we can never point mozart at FABL to aid with partial migrations of things, particularly from morph (envelope responses) to FABL (not).

FABL does support "unwrapped" responses, but requires that a header be passed to trigger that response

FABL docs
- https://github.com/bbc/fabl-modules/tree/main?tab=readme-ov-file#alien-custom-content-types
- https://fabl.api.bbci.co.uk/#usingunwrappedmode

This PR (attempts to) make a change to conditionally add the required header to any requests that are going to a FABL origin, but leave any/all other requests unchanged.

Disclaimer: I've not written any elixir before this, so apologies in advance for the poor implementation! I don't expect this to remain unchanged :)


lib/mozart_fetcher/http_client.ex
```diff
@@ -33,6 +33,14 @@ defmodule HTTPClient do
     end
   end
 
+  defp calculateRequestHeaders(endpoint) do
+    if String.starts_with?(endpoint, "https://fabl.api.") do
+      [{"accept-encoding", "gzip"}, {"ctx-unwrapped", "1"}]
+    else
+      [{"accept-encoding", "gzip"}]
+    end
+  end
```
GuiHeurich, line 42: You could do something like:

`defp request_headers("https://fabl.api."), do: [{"accept-encoding", "gzip"}, {"ctx-unwrapped", "1"}]
defp request_headers(_endpoint), do: [{"accept-encoding", "gzip"}]
`

Then call `request_headers(endpoint)` on line 10.

ettomatic, line 42: Another idiomatic approach is using pattern matching. There's a way to match the beginning of a string (but not the final part as it would be inefficient!)

```elixir
defp request_headers( "https://fabl.api." <> _) do
  [{"accept-encoding", "gzip"}, {"ctx-unwrapped", "1"}]
end

defp request_headers do
  [{"accept-encoding", "gzip"}]
end
```

craigkj: Fast response! Thanks!
This looks better, have changed it to work like this (and Ettore's suggestion below) :)

craigkj:
I also had to update the mix.lock cos `ssl_verify_fun` needed updating. I'm not fully aware of the consequences of this, but I'm assuming this is fairly normal dep bumping?

isavita:
@craigkj could you checkout the `mix.lock` we will prefer not to change on any cost.

ettomatic (approved):

isavita (approved):

JoeARO (approved):

PR #82, opened 2024-06-06
Resframe 5712/upgrade to elixir 1.16.2 otp 26 and erlang 26.2.5

JoeARO:


lib/mozart_fetcher/http_client.ex
```diff
@@ -45,7 +47,7 @@ defmodule HTTPClient do
   end
 
   defp request_headers(headers) do
-    headers ++ [{'User-Agent', 'MozartFetcher'}]
+    headers ++ [{~c"User-Agent", ~c"MozartFetcher"}]
```
isavita, line 50: Shouldn't the header and value be simply string?

lib/mozart_fetcher/http_client.ex
```diff
@@ -1,39 +1,44 @@
...
+      headers = set_request_headers(endpoint)
 
-        options = [
-          recv_timeout: TimeoutParser.parse(endpoint),
-          ssl: MozartFetcher.request_ssl(),
-          hackney: [pool: :origin_pool]
-        ]
+      options = [
+        recv_timeout: TimeoutParser.parse(endpoint),
+        ssl: MozartFetcher.request_ssl() ++ [verify: :verify_none],
```
isavita, line 13: What about adding `[verify: :verify_none]` in `request_ssl` function instead of adding it here, what do you think?

JoeARO: Yea I was thinking about that but then I decided not to, happy to move it in there. It has a test for this function so just a heads up with have a test change too.

JoeARO: Done https://github.com/bbc/mozart-fetcher/pull/82/commits/7d13e1b1931622a223eafe5e8b4f910ef288d2d5

GuiHeurich:
I think you could separate the metrics update from the Elixir/Erlang upgrade into two separate PRs.

JoeARO:
> I think you could separate the metrics update from the Elixir/Erlang upgrade into two separate PRs.

ExMetrics was not allowing the application to start as its pretty out of date, so had to end up doing the metric changes in this PR. I would have probably have left them as were if it weren't for that, as I don't see much point in making changes to these apps. (I probably would have left them on 1.14)

lib/mozart_fetcher.ex
```diff
@@ -25,7 +25,7 @@ defmodule MozartFetcher do
     case environment() do
       :prod -> [certfile: cert_path(), keyfile: key_file_path()]
       _ -> [certfile: System.get_env("DEV_CERT_PEM")]
-    end
+    end ++ [verify: :verify_none]
```
GuiHeurich, line 28: Not sure if it would read better, but do you think we could pattern match on the environment and add `verify none` to both?

lib/mozart_fetcher/fetcher.ex
```diff
@@ -2,34 +2,36 @@ defmodule MozartFetcher.Fetcher do
...
       |> Enum.with_index()
       |> Task.async_stream(&Component.fetch/1, stream_opts)
       |> zip(configs)
       |> decorate_response()
       |> Jason.encode!()
-    end
+
+    timing = (System.monotonic_time(:millisecond) - before_time) |> abs
+    :telemetry.execute([:function, :timing, :fetcher, :process], %{duration: timing})
+    c
```
GuiHeurich, line 34: I almost didn't see the `c` here. I'd add a slightly longer name, just for visibility that you're returning the config after running the metrics.

lib/mozart_fetcher.ex
JoeARO: Ill just add it to both lists, don't think we can pattern match as environment is a function :)

lib/mozart_fetcher/fetcher.ex
JoeARO: Thanks meant to update that but was tired last night!

GuiHeurich (approved):

isavita (approved):

samfrench (approved):

ettomatic (approved):

PR #84, opened 2024-06-19
feat: add ctx-service-env to FABL calls

timasplin:
Added support to set the FABL ctx-service-env using a param that can be passed though from the mozart-page-configs.

Context:

This is for migrating the Shothand module form Morph to FABL.

- We will update the [prestige_content](https://github.com/bbc/mozart-routing/blob/master/src/lib/mozart_routing/payloads/prestige_content.rb) routing to use the `renderer-env` override value that is provided to pass though to the page config.

- The [shorthand config](https://github.com/bbc/mozart-page-configs/blob/ece101b15b0f4daf8db91ca046c16c0af8865a2d/configs/prestige_content/shorthand.json) will then be updated to add `&fabl-ctx-service-env={{renderer-env}}` which will be used by the mozart fetch to see the relevant header.

### DIFF
--- a/lib/mozart_fetcher/http_client.ex
+++ b/lib/mozart_fetcher/http_client.ex
@@ -36,8 +36,10 @@ defmodule HTTPClient do
     end
   end
 
-  defp set_request_headers("https://fabl.api." <> _),
-    do: [{"accept-encoding", "gzip"}, {"ctx-unwrapped", "1"}]
+  defp set_request_headers("https://fabl.api." <> _) do
+    service_env = if MozartFetcher.environment() === :prod , do: "live", else: "test"
+    [{"accept-encoding", "gzip"}, {"ctx-unwrapped", "1"}, {"ctx-service-env", service_env}]
+  end
 
   defp set_request_headers(_endpoint), do: [{"accept-encoding", "gzip"}]
 
--- a/lib/mozart_fetcher/http_client.ex
+++ b/lib/mozart_fetcher/http_client.ex
@@ -7,6 +7,7 @@ defmodule HTTPClient do
     try do
       before_time = System.monotonic_time(:millisecond)
       headers = set_request_headers(endpoint)
+      endpoint = process_request_url(endpoint)
 
       options = [
         recv_timeout: TimeoutParser.parse(endpoint),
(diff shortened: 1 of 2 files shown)


craigkj (approved):

lib/mozart_fetcher/http_client.ex
```diff
@@ -36,8 +36,10 @@ defmodule HTTPClient do
     end
   end
 
-  defp set_request_headers("https://fabl.api." <> _),
-    do: [{"accept-encoding", "gzip"}, {"ctx-unwrapped", "1"}]
+  defp set_request_headers("https://fabl.api." <> _) do
+    service_env = if MozartFetcher.environment() === :prod , do: "live", else: "test"
```
samfrench, line 40: The environment here is "prod" for both test and live environments. I don't think we set an environment variable for test/live in Fetcher but we seem to in other Mozart applications.

ettomatic (approved):

craigkj (approved):

lib/mozart_fetcher/http_client.ex
```diff
@@ -36,8 +37,20 @@ defmodule HTTPClient do
...
+    String.replace(endpoint, ~r"&fabl-ctx-service-env=(test|live)", "")
+  end
+
+  def process_request_url(endpoint) do
+    endpoint
+  end
+
+  defp set_request_headers(endpoint = "https://fabl.api." <> _) do
+    # Default to live to match FABL default behaviour
+    ctx_service_env = if String.match?(endpoint , ~r"fabl-ctx-service-env=test"), do: "test", else: "live"
```
samfrench, line 51: As we need to pass this header to Fabl, should we keep this function matcher to target Fabl, but read the environment from the bake script? It seems brittle to pass in, in the config and assume it is set a certain way. We are then removing this query string above and using it as a header here. It maybe suggests a temporary workaround, but the archive isn't a temporary thing?

ettomatic: So Sam is suggesting to avoid having to parse environments from the page config enad using the Cosmos environment. 
If that's ok for you we can assist in accessing the Runtime enviornment from Cosmos.

PR #88, opened 2024-08-29
Add logic for Fabl headers

samfrench:
https://jira.dev.bbc.co.uk/browse/RESFRAME-5863

### DIFF
--- a/lib/mozart_fetcher/component.ex
+++ b/lib/mozart_fetcher/component.ex
@@ -34,7 +34,11 @@ defmodule MozartFetcher.Component do
   end
 
   defp get(config) do
-    LocalCache.get_or_store(config.endpoint, fn -> HTTPClient.get(config.endpoint) end)
+    headers = config.headers || %{}
+
+    LocalCache.get_or_store(config.endpoint, fn ->
+      HTTPClient.get(config.endpoint, Map.to_list(headers))
+    end)
   end
 
   defp failed_component(component_index, :timeout, id) do
--- a/lib/mozart_fetcher/http_client.ex
+++ b/lib/mozart_fetcher/http_client.ex
@@ -3,10 +3,9 @@ defmodule HTTPClient do
 
   alias MozartFetcher.TimeoutParser
 
-  def get(endpoint, client \\ client()) do
+  def get(endpoint, headers \\ [], client \\ client()) do
     try do
       before_time = System.monotonic_time(:millisecond)
-      headers = set_request_headers(endpoint)
 
       options = [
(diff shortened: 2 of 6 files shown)


ettomatic (approved):

lib/mozart_fetcher/component.ex
```diff
@@ -34,7 +34,11 @@ defmodule MozartFetcher.Component do
   end
 
   defp get(config) do
-    LocalCache.get_or_store(config.endpoint, fn -> HTTPClient.get(config.endpoint) end)
+    headers = config.headers || %{}
+
+    LocalCache.get_or_store(config.endpoint, fn ->
+      HTTPClient.get(config.endpoint, Map.to_list(headers))
```
JoeARO, line 40: Will this work?
I was thinking that we needed a list of Tuples but this would convert it to a list of keyword values.
```elixir
iex(1)> a = %{foo: "bar"}
%{foo: "bar"}
iex(2)> Map.to_list(a)
[foo: "bar"]
iex(3)> b = %{foo: "bar", baz:  "far"}
%{foo: "bar", baz: "far"}
iex(4)> Map.to_list(a)
[foo: "bar"]
iex(5)> Map.to_list(b)
[foo: "bar", baz: "far"]
```

lib/mozart_fetcher/http_client.ex
```diff
@@ -3,10 +3,9 @@ defmodule HTTPClient do
 
   alias MozartFetcher.TimeoutParser
 
-  def get(endpoint, client \\ client()) do
+  def get(endpoint, headers \\ [], client \\ client()) do
```
isavita, line 6: Probably we can add the headers after the client and then will not need to change some other tests or having it called somewhere with second option client. What do you think?

samfrench: The client is only used for mocking in tests. I would prefer to keep the usable parameters together and ideally first.

lib/mozart_fetcher/component.ex
isavita: Even if you have key that are string is returning list of tuples
```elixir
iex(2)> Map.to_list(%{"a" => 42})
[{"a", 42}]
```

isavita (approved):

lib/mozart_fetcher/component.ex
JoeARO: Ahh okay nice, wasn't quite sure so thought I'd double check :)

JoeARO (approved):

craigkj (approved):

PR #94, opened 2025-04-04
Utilise Bake Metadata

JoeARO:


isavita:
@JoeARO 
I think probably don't need the build change in first PR https://github.com/bbc/mozart-fetcher-build/pull/27.
We also probably don't need `bake-scripts/04-create-config-file.sh`.

tracy-o (approved):

samfrench (approved):

PR #98, opened 2025-05-06
Remove debug for config file

JoeARO:


bake-scripts/00-create-config-file.sh
```diff
@@ -58,5 +58,4 @@ echo "  \"secure_configuration\": {}" >> "$CONFIG_JSON"
 echo "}" >> "$CONFIG_JSON"
 
 # cat and echo the JSON file
```
DavidThorpe71, line 60:
```suggestion
# echo the JSON file
```

DavidThorpe71 (commented):
Tiny suggestion, could also remove the comment, whichever you prefer

PR #99, opened 2025-08-20
Replaced Cowboy with Bandit

MikeIrlam:
Updated mix file dependencies
Changed child specs in the app


DavidThorpe71 (commented):
Hi @MikeIrlam  

I think you will need to review the tests a little as some things are failing, looks like the order of some of the fields in now different in some of the test responses.

If you run `mix test` locally you should get some feedback on which ones are failing etc.

Also worth marking these PRs as "Ready for review" so they are no longer "draft" PRs 👍 

Thanks!

JoeARO:
> Hi @MikeIrlam
> 
> I think you will need to review the tests a little as some things are failing, looks like the order of some of the fields in now different in some of the test responses.
> 
> If you run `mix test` locally you should get some feedback on which ones are failing etc.
> 
> Also worth marking these PRs as "Ready for review" so they are no longer "draft" PRs 👍
> 
> Thanks!

We need to add some additional `http_options`before this is ready to review.
Also those tests are flakey I can replicate it on master with: `mix test --seed 906756` I'll create a ticket to address them seperately. I imagine its something to do with Jason.

DavidThorpe71 (approved):

lib/mozart_fetcher/application.ex
```diff
@@ -11,23 +11,17 @@ defmodule MozartFetcher.Application do
   defp children(env: :test) do
     [
       Metrics.TelemetrySupervisor,
-      Plug.Cowboy.child_spec(
-        scheme: :http,
-        plug: MozartFetcher.FakeOrigin,
-        options: [port: 8082]
-      ),
+      {Bandit,
+       plug: MozartFetcher.FakeOrigin, scheme: :http, port: 8082, http_options: [compress: false]},
```
DavidThorpe71, line 15: Just for interest what is the reason for adding the `[compress: false]` option, has a default changed between Cowboy and Bandit?

tracy-o (approved):

samfrench:
Can you also add the ticket to this PR as well.

samfrench (approved):

PR #100, opened 2026-02-24
BELFRAGE-319: Upgraded Elixir & Erlang versions

AWard2022:
- Updated Elixir version to 1.19.5
- Updated Erlang version to 28.3.1
- Updated all dependencies (note: hackney dependency now brought in as a sub dependency of httposion due to conflict)

- Link to working build on Test (using updated Docker image): https://eu-west-1.console.aws.amazon.com/codesuite/codebuild/134209033928/projects/mozart-fetcher/build/mozart-fetcher%3Ab1e67e79-1864-4cfe-ae01-c4a0f618a980/config?region=eu-west-1


.tool-versions
```diff
@@ -1,2 +1,2 @@
-elixir 1.18.3-otp-27
-erlang 27.3.3
\ No newline at end of file
+elixir 1.19.5
```
samfrench: Wouldn't this be 1.19.5-otp-28 ?

samfrench: Does it work for you? I get an error saying it can't find the elixir version. Previously we had the otp specified.

mix.exs
```diff
@@ -6,7 +6,7 @@ defmodule MozartFetcher.MixProject do
     [
       app: :mozart_fetcher,
       version: @version,
-      elixir: "~> 1.18",
+      elixir: "~> 1.19.5",
```
samfrench: Do we specify the patch version? I thought we usually have major and minor as it was before.

mix.exs
```diff
@@ -38,16 +38,15 @@ defmodule MozartFetcher.MixProject do
...
-      {:httpoison, "~> 1.8"},
-      {:con_cache, "~> 1.0"},
-      {:parse_trans, "~> 3.4", override: true},
-      {:hackney, "~> 1.20.1"},
-      {:telemetry, "~> 1.0"},
-      {:telemetry_metrics, "~> 0.6"},
-      {:telemetry_metrics_statsd, "~> 0.7"},
-      {:logger_file_backend, "~> 0.0.10"},
-      {:jason, "~> 1.3"}
+      {:bandit, "~> 1.10.2"},
```
samfrench: Should this be 1.10.3 ?

.tool-versions
AWard2022: Oh right I don't get an error when compiling the project. I can try the otp-28 version

mix.exs
AWard2022: I tried updating the bandit dependency to 1.10.3 however 8 out of 59 tests fail using that version (which I'm currently looking into). Using 1.10.2 doesn't cause any failing tests

mix.exs
AWard2022: Ok I have changed this to the same format as before but with the updated 1.19 version

.tool-versions
AWard2022: There was an error for me with the otp-28 version initially however running asdf install fixed the error and now it works

isavita (approved):

samfrench (approved):

